PR #11, opened 2021-09-21
Add scripts directory for CI builds

bcmendoza:
[ch-6709]

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,6 @@ COPY pkg/ pkg/
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
-RUN go mod tidy
 RUN go mod vendor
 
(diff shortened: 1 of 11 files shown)


Dockerfile
```diff
@@ -5,14 +5,17 @@ WORKDIR /workspace
...
 COPY main.go main.go
 COPY api/ api/
-COPY controllers/ controllers/
+COPY pkg/ pkg/
+
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN go mod download
+RUN go mod tidy
+RUN go mod vendor
```
dontlaugh, line 18: I've found go mod tidy to be problematic inside Docker containers when accessing private repos, but if it's working for you, go for it.

bcmendoza: Eh, It's not actually necessary. Removed and it still builds.

PR #13, opened 2021-09-22
add redis config

moosecanswim:
Add in configuration to operator to specify the redis url to use.


moosecanswim:
[CH 6706]

api/v1alpha1/operatorconfig_types.go
```diff
@@ -31,7 +31,16 @@ type OperatorConfig struct {
...
+	ImagePullSecretName string      `json:"imagePullSecretName"`
+	RedisConfig         redisConfig `json:"redis"`
+	// RedisUrl                   string `json:"redis.url"`
+	// RedisCertificateSecretName string `json:"redis.certificateSecretName"`
+
+}
+
+type redisConfig struct {
+	Url        string `json:"url"`
+	SecretName string `json:"certificateSecretName"`
```
bcmendoza, line 43: Can you move this into the `store` package and export it?

main.go
```diff
@@ -53,13 +53,18 @@ func main() {
 	var probeAddr string
 	var enableLeaderElection bool
 	var development bool
+	var redisUrl string
```
bcmendoza, line 56: We don't need command line args for each option -- just the ones the controller-runtime manager requires, plus a `development` flag for convenience in local development. You can remove lines 56-57 and 65-66.

main.go
```diff
@@ -97,6 +102,17 @@ func main() {
...
+		opConfig.RedisConfig.Url = "redis://greymatteroperator:redis@localhost:6379/0" // TODO: if this field is empty and we will be spinning up a redis instance then it should connect to it <svc name>.<current namespace>.svc.cluster.local
+	}
+	// TODO: if certificateSecretName exists then use tls to connect to redis (https://redis.uptrace.dev/guide/server.html#using-tls)
+	if opConfig.RedisConfig.SecretName != "" {
+		// get secret specified
+		// validate there are 'ca', 'server_cert', 'server_key' keys.
+		// set up redis to use tls
+		setupLog.Info("A redis config secret name was provided.  TODO: Need to set up redis tls")
+	}
+
```
bcmendoza, line 115: Let's save this initialization logic for when we call `store.New()`. Maybe move this into `store.go` in a separate function for now?

bcmendoza (changes requested):

main.go
moosecanswim: done.  also added testing.

PR #14, opened 2021-09-23
Add InstallationConfig custom resource

bcmendoza:
Starts [ch-6700](https://app.shortcut.com/grey-matter/story/6700/write-gmcore-package).

This PR adds an `InstallationConfig` custom resource for defining "values" templates that will eventually be used to install Grey Matter core components.

The `gmcore` package now embeds a filesystem using the `go:embed` compiler directive which allows us to version these "values" files for each Grey Matter version. A later improvement will involve allowing this CR to be modified via the apiserver, enabling users to seamlessly upgrade deployments the same way `helm upgrade` does. [Here's an example "values" file](https://github.com/greymatter-io/operator/blob/3e0aafc6afaf9cf3480ba131f9c8887e2b45aec9/pkg/gmcore/values/v1.6.yaml).

I've also started to define options for the `InstallationConfig` using the functional opts pattern for adding "overlays" to base configs. For example, given a `Mesh` CR value, Operator will call `SystemValuesConfig.Overlay(WithSPIRE)`, which will install SPIRE for the given mesh and inject SPIRE configs into all proxy nodes.


moosecanswim (approved):
LGTM

PR #17, opened 2021-09-24
Change Mesh CR scope to namespace

bcmendoza:
Mesh custom resources should be created in a namespace, the same namespace where the core services will be installed.

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -9,6 +9,7 @@ repo: github.com/greymatter-io/operator
 resources:
 - api:
     crdVersion: v1
+    namespaced: true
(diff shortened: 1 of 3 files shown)


bcmendoza:
Withdrawing this. It's making a bit more sense to me now that we should keep the Mesh CR at the cluster scope. Resources that have a `namespaced name` are those that can be duplicated across namespaces (i.e. with the same `name`), but that's confusing within our Operator code and also for the user. I'd rather not have a bunch of `zone-default-zone` Mesh objects in different namespaces. Let's keep the Mesh CR at the cluster scope.

PR #19, opened 2021-09-28
Redis system values config

moosecanswim:


api/v1alpha1/installvalues.go
```diff
@@ -78,8 +87,93 @@ func SPIRE(installValues *InstallValues) {
 
 // A InstallValues option that injects configuration for a Redis provider.
 // If the Redis configuration is empty, adds Values for configuring an internal Redis.
-func Redis(redisHost, redisPort string) func(*InstallValues) {
+func Redis(rc RedisConfig, msh Mesh) func(*InstallValues) {
```
bcmendoza, line 90: Looks like you only need the `namespace` string as a parameter vs the entire Mesh object.

api/v1alpha1/installvalues.go
```diff
@@ -78,8 +87,93 @@ func SPIRE(installValues *InstallValues) {
...
+
+	// If redisConfig is nil then we need to add in redis values to create a redis deployment
+	// Otherwise we need to
+
 	return func(installValues *InstallValues) {
-		// inject configs into install values for control api, catalog, jwt security
+
+		//Initialize default redis values to be reasigned based on redis config
+		redis_host := ""
```
bcmendoza, line 98: Use camelCase variables in Go

api/v1alpha1/installvalues.go
```diff
@@ -78,8 +87,93 @@ func SPIRE(installValues *InstallValues) {
...
+			redis_port = "6379"
+			redis_port_int, _ := strconv.Atoi(redis_port)
+			redis_password = "redis"
+			redis_db = "0"
+
+			// Add redis values if we need to create one
+			installValues.Redis.With(
+				Image("bitnami/redis:5.0.12"),
+				Command("redis-server"),
+				Args([]string{"--appendonly", "yes", "--requirepass", "$(REDIS_PASSWORD)", "--dir", "/data"}),
```
bcmendoza, line 116: Add options for `username` and `password` to the `RedisConfig` so that you can pass them in here.

api/v1alpha1/installvalues.go
```diff
@@ -78,8 +87,93 @@ func SPIRE(installValues *InstallValues) {
...
+		redis_port := "6379"
+		redis_password := "redis" //TODO: randomize this internal only password
+		redis_db := "0"
+
+		if rc.Url == "" {
+
+			svc_name := "redis"
+			redis_host = fmt.Sprintf("%s.%s.svc.cluster.local", svc_name, msh.Namespace)
+			redis_port = "6379"
+			redis_port_int, _ := strconv.Atoi(redis_port)
```
bcmendoza, line 108: Unsafe; but also, since you already know it will be 6379, might as well define it i.e. `redisPortInt := int32(6379)`.

api/v1alpha1/installvalues.go
```diff
@@ -78,8 +87,93 @@ func SPIRE(installValues *InstallValues) {
...
+			// Add redis values if we need to create one
+			installValues.Redis.With(
+				Image("bitnami/redis:5.0.12"),
+				Command("redis-server"),
+				Args([]string{"--appendonly", "yes", "--requirepass", "$(REDIS_PASSWORD)", "--dir", "/data"}),
+				Envs(map[string]string{
+					"REDIS_PASSWORD": redis_password,
+				}),
+				Port("internal-greymatter-redis", corev1.ContainerPort{
+					Name:          svc_name,
```
bcmendoza, line 121: `Name` should not be set here, as it will later on be assigned the first argument in the `Port` option (i.e. `"internal-greymatter-redis"`).

api/v1alpha1/installvalues.go
```diff
@@ -78,8 +87,93 @@ func SPIRE(installValues *InstallValues) {
...
+				//TODO: add /data volume and volume mount
+			)
+
+			// add volume and volume mount if tls secret exists
+
+		} else {
+			//Parse given redis url and assign values to previously initialized variables
+			u, err := url.Parse(rc.Url)
+			if err != nil {
+				panic(err)
```
bcmendoza, line 143: We'll probably need to do validation for the `RedisConfig` prior to passing it to the `Redis` option. Can you just add a TODO here to handle this elsewhere and eventually log it here?

api/v1alpha1/installvalues.go
```diff
@@ -78,8 +87,93 @@ func SPIRE(installValues *InstallValues) {
...
+			}
+			redis_host, redis_port, _ = net.SplitHostPort(u.Host)
+			redis_password = "redis"
```
bcmendoza, line 146: ?

api/v1alpha1/mesh_types.go
```diff
@@ -23,7 +23,8 @@ import (
 // MeshSpec defines the desired state of Mesh
 type MeshSpec struct {
 	// Foo is an example field of Mesh. Note that json tags are required for fields to be serialized.
-	Foo string `json:"foo,omitempty"`
+	Foo         string       `json:"foo,omitempty"`
```
bcmendoza, line 26: Can you remove `Foo` here and then run `make generate manifests`?

api/v1alpha1/installvalues.go
```diff
@@ -78,8 +87,93 @@ func SPIRE(installValues *InstallValues) {
...
-func Redis(redisHost, redisPort string) func(*InstallValues) {
+func Redis(rc RedisConfig, msh Mesh) func(*InstallValues) {
+
+	// If redisConfig is nil then we need to add in redis values to create a redis deployment
+	// Otherwise we need to
+
 	return func(installValues *InstallValues) {
-		// inject configs into install values for control api, catalog, jwt security
+
+		//Initialize default redis values to be reasigned based on redis config
```
bcmendoza, line 97: The defaults should already be set in `installValues` (via our embedded YAML files). If `RedisConfig` is not nil, make `installValues.Redis` nil so that `gmcore.Installer` will know not to create an internal Redis deployment.

api/v1alpha1/installvalues.go
```diff
@@ -78,8 +87,93 @@ func SPIRE(installValues *InstallValues) {
...
+			redis_host, redis_port, _ = net.SplitHostPort(u.Host)
+			redis_password = "redis"
+			redis_db = strings.ReplaceAll(u.Path, "/", "")
+		}
+
+		//modify controlapi values
+		installValues.ControlAPI.With(
+			Envs(map[string]string{
+				"GM_CONTROL_API_REDIS_HOST": redis_host,
+				"GM_CONTROL_API_REDIS_PORT": string(redis_port),
```
bcmendoza, line 154: `redis_port` is already a string

bcmendoza (changes requested):
Good work so far. Requested some changes. When you write tests, you should read in defaults from `fixture.yaml` for `installValues.Redis`, and when you check for the end result after applying the `Redis` option, `installValues.Redis` should be set to `nil`.

api/v1alpha1/installvalues.go
moosecanswim: wait what i said was dumb https://www.google.com/search?client=firefox-b-1-d&q=int32+max+value

api/v1alpha1/installvalues.go
moosecanswim: Ah gotcha.

api/v1alpha1/installvalues.go
moosecanswim: The `url.Parse()` function return `u.Host` which includes host and port https://gobyexample.com/url-parsing

bcmendoza: ah, this comment was meant to be about assigning `redis` as the `redis_password`

api/v1alpha1/installvalues.go
bcmendoza: Actually just kidding. These are for the internal deployment, so it should be fine to just generate and use a password here.

api/v1alpha1/mesh_types.go
```diff
@@ -51,6 +50,14 @@ type MeshList struct {
 	Items           []Mesh `json:"items"`
 }
 
+// RedisConfig contains the redis connection information for a given mesh installation
+type RedisConfig struct {
+	Url        string `json:"url"`
+	Username   string `json:"username"`
+	Password   string `json:"password"`
```
bcmendoza, line 57: @moosecanswim Sorry about my earlier comment about adding the `username` and `password` here. These should be added directly to the `url` as a properly formatted Redis connection string.

bcmendoza: (meaning they can be removed from RedisConfig)

bcmendoza (approved):
LGTM

PR #23, opened 2021-10-06
add watch namespaces and ensure that the values in the envar are uniq…

moosecanswim:
…ue; adds tests too

### DIFF
--- a/api/v1alpha1/mesh_types.go
+++ b/api/v1alpha1/mesh_types.go
@@ -36,6 +36,9 @@ type MeshSpec struct {
 	// +kubebuilder:default=10808
 	// +optional
 	MeshPort int32 `json:"mesh_port"`
+	// The other namespaces the control shouls monitor (Should not include the namespace the mesh lives in)
+	// +kubebuilder:validation:Pattern=`([\S]+,)*([\S]+)`
+	WatchNamespaces string `json:"watch_namespaces",omitempty`
 	// WatchNamespaces []string
 }
 
--- a/pkg/version/cue.mod/inputs.cue
+++ b/pkg/version/cue.mod/inputs.cue
@@ -12,3 +12,5 @@ Redis: {
   password: "" | string
   db: *"0" | string
 }
+
+WatchNamespaces : string
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -115,3 +115,32 @@ func Redis(externalURL string) InstallOption {
 		)
 	}
 }
+
+// An InstallOption for injection watch namespaces
+func WatchNamespaces(meshNamespace string, watchNamespaces string) InstallOption {
+	return func(v *Version) {
+		var wn []string
+		if watchNamespaces != "" {
+			wn = strings.Split(watchNamespaces, ",")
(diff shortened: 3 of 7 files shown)


api/v1alpha1/mesh_types.go
```diff
@@ -36,6 +36,9 @@ type MeshSpec struct {
 	// +kubebuilder:default=10808
 	// +optional
 	MeshPort int32 `json:"mesh_port"`
+	// The other namespaces the control shouls monitor (Should not include the namespace the mesh lives in)
```
bcmendoza, line 39:
```suggestion
	// Namespaces the mesh data plane should include in its network (in addition to the namespace where the Mesh control plane is being installed)
```

pkg/version/cue.mod/inputs.cue
```diff
@@ -12,3 +12,5 @@ Redis: {
   password: "" | string
   db: *"0" | string
 }
+
+WatchNamespaces : string
```
bcmendoza, line 16:
```suggestion
WatchNamespaces: string

```

pkg/version/version.go
```diff
@@ -115,3 +115,32 @@ func Redis(externalURL string) InstallOption {
 		)
 	}
 }
+
+// An InstallOption for injection watch namespaces
```
bcmendoza, line 119:
```suggestion
// An InstallOption for injecting watch namespaces
```

pkg/version/version.go
```diff
@@ -115,3 +115,32 @@ func Redis(externalURL string) InstallOption {
...
+
+func unique(slice []string) []string {
+	keys := make(map[string]bool)
+	list := []string{}
+	for _, entry := range slice {
+		if _, value := keys[entry]; !value {
+			keys[entry] = true
+			list = append(list, entry)
+		}
+	}
```
bcmendoza, line 144: This is a idiomatic way to convert a list into a set:
```go
keys := make(map[string]struct{})
for _, entry := range slice {
    keys[entry] = struct{}{}
}
var list []string
for k := range keys {
    list = append(list, k)
}
```

I think it'd be nice if we did this in Cue though... https://cuetorials.com/patterns/unique-list/

pkg/version/versions_test.go
```diff
@@ -172,6 +173,92 @@ func TestVersions(t *testing.T) {
...
+						if envarValue == "" {
+							t.Fatal("Environment Variable [GM_CONTROL_KUBERNETES_NAMESPACES] has no value specified in Control container")
+						}
+
+						expSlice := strings.Split(expectedValue, ",")
+						envarValueSlice := strings.Split(envarValue, ",")
+						// envar is not as expected
+						if stringSlicesEqual(expSlice, envarValueSlice) {
+							t.Fatalf("Environment Variable [GM_CONTROL_KUBERNETES_NAMESPACES] actual: [%s] expected: [%s]", envarValue, expectedValue)
+						}
```
bcmendoza, line 254: I don't think we need to test for structural equality here. Just check the strings for equality since that's what was set in the Cue value.

PR #26, opened 2021-10-14
Adds the ability to create a route or an ingress to the cluster when a new mesh is created

moosecanswim:
This adds an additional field to the crd that defines a mesh, adds options to the version and then parses that to create an ingress or route depending on the cluster deployed.


main.go
```diff
@@ -58,6 +61,7 @@ func init() {
 //+kubebuilder:rbac:groups=apps,resources=deployments;statefulsets,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=services;configmaps;serviceaccounts;secrets;pods,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=route.openshift.io,resources=routes,verbs=get;list;watch;create;update;patch;delete
```
bcmendoza, line 64: Should there also be an RBAC rule for ingress? (extensions)

pkg/installer/install.go
```diff
@@ -186,7 +191,7 @@ func createOrUpdate(ctx context.Context, c client.Client, obj client.Object) (st
 		return "create", "success", nil
 	}
 
-	if err := c.Update(ctx, obj); err != nil {
+	if err := c.Update(ctx, existing.(client.Object)); err != nil {
```
bcmendoza, line 194: I don’t think we want to do this. I deep copied the object so that the overwrite from Get would modify “existing”. To call Update with “existing” is simply sending the apiserver the same object retrieved via Get.

pkg/version/version.go
```diff
@@ -151,3 +151,29 @@ func JWTSecrets(v *Version) {
...
+	clusterType = strings.ToLower(clusterType)
+	return func(v *Version) {
+		v.cue = v.cue.Unify(Cue(fmt.Sprintf(`ClusterType: "%s"`, clusterType)))
+	}
+}
+
+func Domain(addr string) InstallOption {
+	return func(v *Version) {
+		var domain string
+		clusterType := fmt.Sprintf("%s", v.cue.LookupPath(cue.ParsePath("ClusterType")))
```
bcmendoza, line 167: Nice!! Way to dig into the Cue API. Is there any benefit to separating the options? Maybe it’d be better to just use ClusterType and Domain together in one option.

pkg/version/versions_test.go
```diff
@@ -5,7 +5,9 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/ghodss/yaml"
+	// "github.com/ghodss/yaml"
+
+	"gopkg.in/yaml.v2"
```
bcmendoza, line 10: "gopkg.in/yaml” is the devil. It doesn’t parse deep nested JSON. Use “ghodss/yaml" pls!

api/v1alpha1/mesh_types.go
```diff
@@ -24,6 +24,13 @@ import (
 
 // Defines the desired state of a Grey Matter mesh.
 type MeshSpec struct {
+	// The type of cluster the mesh is installed into
+	// +kubebuilder:validation:Enum="openshift";"kubernetes"
+	// +kubebuilder:default="openshift"
+	ClusterType string `json:"cluster_type"`
```
bcmendoza, line 30: Hmm. Now that I think about it, this should be in the BootstrapConfig. And really, it could be something we discover at startup.

pkg/version/cue.mod/inputs.cue
```diff
@@ -5,6 +5,9 @@ InstallNamespace: string
 // The scope of the mesh network; includes InstallNamespace
 WatchNamespaces: string
 
+ClusterType: *"openshift" | string
+Domain: *"localhost" | string
```
bcmendoza, line 9: Could we go with IngressDomain here? It’d be easy to confuse this with our Domain mesh object.

pkg/installer/install.go
```diff
@@ -134,6 +134,11 @@ MANIFEST_LOOP:
 		if group.StatefulSet != nil {
 			i.apply(group.StatefulSet, mesh, scheme)
 		}
+		if group.Route != nil {
+			i.apply(group.Route, mesh, scheme)
+		} else if group.Ingress != nil {
```
bcmendoza, line 139: You can omit the else here since the logic is set in Cue so that only one of these resources should ever be available.

pkg/version/version.go
```diff
@@ -151,3 +151,29 @@ func JWTSecrets(v *Version) {
...
+		switch clusterType {
+		case "openshift":
+			// openshift will assign a host internally
+			// domain = fmt.Sprintf("%s.%s.%s", "edge", installNamespace, addr)
+		case "kubernetes":
+			domain = addr
+		}
+		domain = strings.TrimSpace(domain)
+		fmt.Printf("\n\n----\ndomain: %s\n----\n", domain)
```
bcmendoza, line 176: This area needs some cleanup

pkg/version/versions_test.go
```diff
@@ -44,6 +46,7 @@ func TestVersions(t *testing.T) {
 				options        []InstallOption
 				checkManifests func(*testing.T, []ManifestGroup)
 				checkSidecar   func(*testing.T, Sidecar)
+				// checkIngress   func(*testing.T, extv1b1.Ingress)
```
bcmendoza, line 49: Cleanup

bcmendoza (changes requested):

pkg/version/version.go
moosecanswim: Now that clusterType is set from the operator config doing it separately makes more sense imo

pkg/installer/install.go
moosecanswim: So this is the error i get when changing it back to obj
```
t2","Route":"mesh-sample","error":"routes.route.openshift.io \"mesh-sample\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update"}
```

pkg/version/cue.mod/outputs.cue
```diff
@@ -3,6 +3,8 @@ package base
 import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+  routev1 "github.com/openshift/api/route/v1"
+  extv1beta1 "k8s.io/api/extensions/v1beta1"
```
bcmendoza, line 7: This should be formatted...

pkg/version/cue.mod/outputs.cue
```diff
@@ -186,6 +188,63 @@ manifests: [...#ManifestGroup] & [
...
+    route: routev1.#Route & {  
+      {
+        apiVersion: "route.openshift.io/v1"
+        kind: "Route"
+        metadata: {
+          name: MeshName
+          namespace: InstallNamespace
+        }
+        spec: {
+          port: targetPort: "proxy"
```
bcmendoza, line 201: You should be able to add `subdomain: MeshName` here, and the host value will be populated as `<meshname>.apps.mycluster.com`.

bcmendoza: You can get the domain name of the OpenShift cluster via the cluster-scoped resource `config.openshift.io/v1.Ingress`:
```
apiVersion: config.openshift.io/v1
kind: Ingress
metadata:
  name: cluster
spec:
  domain: apps.coffee.greymatter.services
```

moosecanswim: I figure we keep the route and ingress separate for now atleast.  That functionality was set up for cases where the app stack does not have the ability to create a route so that is a work around for them.  We can so i figure we should.

pkg/installer/install.go
moosecanswim: Fixed

pkg/version/cue.mod/outputs.cue
moosecanswim: Created a separate issue for this

pkg/version/cue.mod/outputs.cue
```diff
@@ -184,6 +186,63 @@ manifests: [...#ManifestGroup] & [
...
+            }
+            wildcardPolicy: "None"
+          }
+        }
+      }
+    }
+    if ClusterType == "kubernetes" {
+      ingress: extv1beta1.#Ingress & {
+        {
+          apiVersion: "extensions/v1beta1"
```
bcmendoza, line 214: `apiVersion: "networking.k8s.io/v1beta1"`

pkg/version/cue.mod/outputs.cue
```diff
@@ -184,6 +186,63 @@ manifests: [...#ManifestGroup] & [
       }
     }
   ]
+  if _c[0].name == "edge" {
+    if ClusterType == "openshift" {
+      route: routev1.#Route & {
```
bcmendoza, line 191: [Refer to this doc](https://docs.openshift.com/container-platform/4.8/networking/routes/route-configuration.html#nw-ingress-creating-a-route-via-an-ingress_route-configuration) which says that creating an explicit route isn't necessary. We can just create an Ingress object and let Openshift create the route for us. This should help simplify this PR some.

pkg/version/cue.mod/outputs.cue
```diff
@@ -184,6 +186,63 @@ manifests: [...#ManifestGroup] & [
...
+          apiVersion: "extensions/v1beta1"
+          kind: "Ingress"
+          metadata: {
+            name: "greymatter-ingress"
+            namespace: InstallNamespace
+            annotations: {
+              "nginx.ingress.kubernetes.io/ssl-passthrough": "true"
+              "nginx.ingress.kubernetes.io/force-ssl-redirect": "true"
+              "nginx.ingress.kubernetes.io/backend-protocol": "https"
+            }
```
bcmendoza, line 223: These should only be set if TLS is configured, right?

pkg/version/cue.mod/outputs.cue
```diff
@@ -184,6 +186,63 @@ manifests: [...#ManifestGroup] & [
...
+          }
+          spec: {
+            rules: [
+              {
+                host: IngressSubDomain
```
bcmendoza, line 228: I think this should be `host: "\(MeshName).\(IngressSubDomain)"`

main.go
```diff
@@ -47,7 +49,8 @@ var (
 
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
-
+	utilruntime.Must(routev1.AddToScheme(scheme))
+	utilruntime.Must(extv1beta1.AddToScheme(scheme))
```
bcmendoza, line 53: I don't think `extensions/v1beta1` needs to be added to the controller scheme since it is a Kubernetes resource group. Those are added automatically.

pkg/version/cue.mod/outputs.cue
moosecanswim: I feel like we want to keep this as is because eventually we'll validate the proposed route against all routes in the cluster to ensure its unique.

pkg/version/cue.mod/outputs.cue
```diff
@@ -184,6 +185,42 @@ manifests: [...#ManifestGroup] & [
       }
     }
   ]
+  if _c[0].name == "edge" {
+      ingress: netv1b1.#Ingress & {
+        {
+          apiVersion: "networking.k8s.io/v1beta1"
```
bcmendoza, line 191: Logs show this: `{"level":"info","ts":1634590499.7205205,"logger":"KubeAPIWarningLogger","msg":"networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress"}`. Apparently we should be using `v1`.

pkg/version/version.go
```diff
@@ -163,3 +163,18 @@ func JWTSecrets(v *Version) {
 		}`,
 	))
 }
+
+func IngressSubDomain(addr string) InstallOption {
+	return func(v *Version) {
+		installNamespace := fmt.Sprintf("%s", v.cue.LookupPath(cue.ParsePath("InstallNamespace")))
+		meshName := fmt.Sprintf("%s", v.cue.LookupPath(cue.ParsePath("MeshName")))
+		// TODO: validate mesh name is unique.  once this is done we can strip the namespace from the subdomain spec
+		v.cue = v.cue.Unify(cueutils.FromStrings(fmt.Sprintf(`IngressSubDomain: "%s.%s.%s"`, strings.TrimSpace(meshName), strings.TrimSpace(installNamespace), strings.TrimSpace(addr))))
```
bcmendoza, line 172: Let's assume that we're gonna tackle that validation sooner than later, and go ahead and remove the installNamespace from this.

config/manager/bootstrap_config.yaml
```diff
@@ -9,3 +9,4 @@ leaderElection:
   leaderElect: true
   resourceName: 715805a0.greymatter.io
 imagePullSecretName: gm-docker-secret
+clusterType: openshift
```
bcmendoza, line 12: This is no longer needed

pkg/bootstrap/bootstrapconfig_types.go
```diff
@@ -31,4 +31,6 @@ type BootstrapConfig struct {
 	// The name of the secret in the namespace where Grey Matter Operator is deployed.
 	// This secret is re-created in each namespace where Grey Matter Core is installed.
 	ImagePullSecretName string `json:"imagePullSecretName"`
+	// This is the type of cluster the operator is installed into (openshift or kubernetes)
```
bcmendoza, line 34: Not needed anymore

main.go
```diff
@@ -132,8 +134,14 @@ func main() {
 		cfg.ImagePullSecretName = "gm-docker-secret"
 	}
 
+	// Set the default cluster type to openshift
+	// TODO: Make the operator automatically determine which clusterType it is running in
+	if cfg.ClusterType == "" {
```
bcmendoza, line 139: Remove

bcmendoza (changes requested):

pkg/version/cue.mod/inputs.cue
```diff
@@ -7,11 +7,16 @@ InstallNamespace: string
 // The scope of the mesh network; includes InstallNamespace
 WatchNamespaces: string
 
+MeshName: string
```
bcmendoza, line 10: This is a duplicate

pkg/version/cue.mod/outputs.cue
```diff
@@ -1,8 +1,9 @@
 package base
 
 import (
-	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
+	appsv1     "k8s.io/api/apps/v1"
+	corev1     "k8s.io/api/core/v1"
+  netv1      "k8s.io/api/networking/v1"
```
bcmendoza, line 6: go fmt

pkg/version/cue.mod/outputs.cue
```diff
@@ -184,6 +185,44 @@ manifests: [...#ManifestGroup] & [
...
+          apiVersion: "networking.k8s.io/v1"
+          kind: "Ingress"
+          metadata: {
+            name: "greymatter-ingress"
+            namespace: InstallNamespace
+            if EdgeTlsIngress {
+              annotations: {
+              "nginx.ingress.kubernetes.io/ssl-passthrough": "true"
+              "nginx.ingress.kubernetes.io/force-ssl-redirect": "true"
+              "nginx.ingress.kubernetes.io/backend-protocol": "https"
```
bcmendoza, line 200: indent

bcmendoza (approved):

PR #30, opened 2021-10-21
Support vanilla k8s and deploy to kOps

alecholmez:
Signed-off-by: Alec Holmes <alecholmez@me.com> and Brian Mendoza.

This PR adds support for deploying to Kubernetes. The operator has been deployed to a kOps cluster at `gm-operator-dev.k8s.local`, with mesh components installed in the `default` namespace. Note that ingress currently does not work since that is being addressed in a separate PR. For now, you can port-forward to edge at 10808.

To support deploying in Kubernetes without OpenShift/OLM support, this PR:
- Adds scripts that use cfssl to generate a CA and cert/key pair for the operator webhooks
- Cleans up some unused kustomize files that we don't plan to use, i.e. cert-manager
- Makes a few tweaks to get things working in kOps, i..e `securityContext` added to Redis with `fsGroup: 2000`

The README has been updated and is a good guide to follow for testing this PR.


pkg/version/cue.mod/outputs.cue
```diff
@@ -54,6 +54,11 @@ manifests: [...#ManifestGroup] & [
               { name: "gm-docker-secret" }
             ]
           }
+          if _c[0].name == "gm-redis" || _c[0].name == "gm-prometheus" {
+            securityContext: {
+              fsGroup: 2000
```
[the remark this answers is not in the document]
bcmendoza: Good call. We’ll need to get your PR in first then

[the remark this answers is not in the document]
moosecanswim: kk

main.go
```diff
@@ -89,10 +90,10 @@ func main() {
 	// These values will not be replaced by any values set in a read configFile.
 	var err error
 	options := ctrl.Options{
-		Scheme:         scheme,
-		LeaderElection: enableLeaderElection,
-		// TODO: Generate hash for ID; should be unique with multiple operator replicas
-		LeaderElectionID: "715805a0.greymatter.io",
+		Scheme:                  scheme,
+		LeaderElection:          enableLeaderElection,
+		LeaderElectionID:        "715805a0.greymatter.io",
```
dontlaugh, line 95: Should we leave the TODO in

pkg/webhooks/certs.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"bytes"
+	"html/template"
+	"strings"
+
+	"github.com/Masterminds/sprig"
+)
+
+// NOTE: This code does not run, but is retained here for reference and to use later for generating certs.
```
dontlaugh, line 11: Does "does not run" mean "is currently unused"?

dontlaugh (commented):
Looks good. I still need to test.

pkg/webhooks/certs.go
bcmendoza: Yes, it is currently unused. I didn't wanna get rid of this code right away since we may have some use for it later on.

main.go
bcmendoza: Yes... I'm not quite sure why kubebuilder has this hard-coded in just yet, or if we even need leader election since there should only ever be one pod that handles requests.

dontlaugh: Not a blocker for this PR

bcmendoza:
To test with an existing kOps cluster:
1. Ensure you have `NEXUS_USER` and `NEXUS_PASSWORD` sourced in your environment with Docker credentials for pulling Grey Matter images.
2. Run `./scripts/dev deploy k8s` to create all resources used to run the operator, as well as the operator itself
3. You may need to wait a few seconds after the operator is deployed for its server to be launched
4. Run `./scripts/dev sample k8s` to create a Mesh custom resource in the `default` namespace.
5. Watch the log outputs for installing mesh components/dependencies and configuring the mesh
6. To clean up, run `./scripts/dev cleanup k8s`.

scripts/dev
```diff
@@ -2,79 +2,189 @@
...
+
+k3d_sample() {
+  kubectl apply -f hack/sample-k8s.yaml
+  k3d_logs
+}
+
+# Shared
+
+prompt_push() {
+  read -p "Build and push a new Docker image? " -n 1 -r
```
dontlaugh, line 115: Add (y/n) to this prompt

bcmendoza:
@dontlaugh This project bundles CLI 3.0 for now and expects it. Later when we move to 4.0, we'll bundle that and support it. I'd rather just `t.Skip` here in the meantime, since the operator should only be running in-cluster, and with the right CLI bundled.

dontlaugh (changes requested):
Make scripts/bootstrap complain if `kustomize` is missing

dontlaugh:
After the kustomize thing, this is mergable

dontlaugh (approved):

PR #32, opened 2021-10-21
Multi tennent

moosecanswim:
This Pr promotes the mesh custom resource to be scoped to the cluster level.  This pr also introduces validation to ensure that the `install_namespace` is unique amongst all mesh resources in the cluster.

To test this attempt to create two mesh objects with the same install_namespace configured.

Tests:

Create a mesh: `oc apply -f <this file>`

```
apiVersion: greymatter.io/v1alpha1
kind: Mesh
metadata:
  name: mesh-sample
spec:
  cluster_url: apps.coffee.greymatter.services
  edge_tls_ingress: false
  mesh_port: 10808
  release_version: '1.6'
  zone: default-zone
  install_namespace: gm-operator-test2
```

Next create another mesh: `oc apply -f <this file>`

```
apiVersion: greymatter.io/v1alpha1
kind: Mesh
metadata:
  name: mesh-sample2
spec:
  cluster_url: apps.coffee.greymatter.services
  edge_tls_ingress: false
  mesh_port: 10808
  release_version: '1.6'
  zone: default-zone
  install_namespace: gm-operator-test2
```

The mesh creation should fail with the following error message 

```
Error from server (A mesh exists with the specified install_namespace already.): error when creating "gm-operator-test3.yaml": admission webhook "validate-mesh.greymatter.io" denied the request: A mesh exists with the specified install_namespace already.
```


pkg/webhooks/mesh.go
```diff
@@ -65,6 +67,26 @@ func (mv *meshValidator) Handle(ctx context.Context, req admission.Request) admi
...
+		if err := mv.DecodeRaw(req.Object, obj); err != nil {
+			return admission.Errored(http.StatusInternalServerError, err)
+		}
+		if mesh.Spec.InstallNamespace == obj.Spec.InstallNamespace {
+			hasInstallNamespace = true
+		}
+	}
+	if hasInstallNamespace {
+		return admission.ValidationResponse(false, "A mesh exists with the specified install_namespace already.")
+	}
```
bcmendoza, line 88: You can exit early on line 83 rather than track a `hasInstallNamespace` variable.

pkg/webhooks/mesh.go
```diff
@@ -65,6 +67,26 @@ func (mv *meshValidator) Handle(ctx context.Context, req admission.Request) admi
...
+
 	// TODO: Ensure only one mesh exists in a namespace
```
bcmendoza, line 90: Done! Delete this

pkg/webhooks/webhooks.go
```diff
@@ -6,15 +6,16 @@ import (
...
 
 var (
 	logger = ctrl.Log.WithName("webhooks")
 )
 
-func Register(mgr ctrl.Manager, i *installer.Installer, c *cli.CLI) {
+func Register(mgr ctrl.Manager, i *installer.Installer, c *cli.CLI, ctrlClient ctrlclient.Client) {
 	mgr.GetWebhookServer().Register("/mutate-mesh", &admission.Webhook{Handler: &meshDefaulter{Installer: i}})
-	mgr.GetWebhookServer().Register("/validate-mesh", &admission.Webhook{Handler: &meshValidator{Installer: i, CLI: c}})
+	mgr.GetWebhookServer().Register("/validate-mesh", &admission.Webhook{Handler: &meshValidator{Installer: i, CLI: c, Client: ctrlClient}})
```
bcmendoza, line 19: In keeping with convention, you can just say `cc` or `cl` here. Go encourages short variable names.

pkg/webhooks/mesh.go
```diff
@@ -65,6 +67,26 @@ func (mv *meshValidator) Handle(ctx context.Context, req admission.Request) admi
...
+		return admission.ValidationResponse(false, "Unable to get list of existing mesh resources")
+	}
+	var hasInstallNamespace bool
+	// parse through meshes  and see if the request object Namespace is already in a mesh spec
+	for _, mesh := range meshList.Items {
+		obj := &v1alpha1.Mesh{}
+		if err := mv.DecodeRaw(req.Object, obj); err != nil {
+			return admission.Errored(http.StatusInternalServerError, err)
+		}
+		if mesh.Spec.InstallNamespace == obj.Spec.InstallNamespace {
```
bcmendoza, line 82: You shouldn't be decoding `req.Object` here into `obj` inside of this loop, or at all. Instead, you can compare `mesh.Spec.Namespace` with `req.Namespace`. The `req.Namespace` will be the same namespace as the Mesh CR being validated.

moosecanswim: `req.Namespace` was comming up as blank when i was debugging this initially.

pkg/webhooks/mesh.go
```diff
@@ -65,6 +67,22 @@ func (mv *meshValidator) Handle(ctx context.Context, req admission.Request) admi
 		return admission.ValidationResponse(false, "attempted to create Mesh in 'gm-operator' namespace")
 	}
 
+	// Ensure that each mesh spec has a unique install-namespace
+	meshList := &v1alpha1.MeshList{}
+	if err := mv.List(context.TODO(), meshList); err != nil {
+		return admission.ValidationResponse(false, "Unable to get list of existing mesh resources")
+	}
+	// parse through meshes  and see if the request object Namespace is already in a mesh spec
+	obj := &v1alpha1.Mesh{}
```
bcmendoza, line 76: You can just use the code from line 90 and on. Since the mesh is decoded there, it shouldn’t be re-decoded further down.

bcmendoza (changes requested):

PR #36, opened 2021-11-05
WIP: greymatter.io/network-mode label

moosecanswim:
This pr adds the ability to add the annotation `greymatter.io/exclude-from-mesh: "true"` to deployments and stateful sets and no sidecar will be added and mesh configs wont be created.

To test:
- deploy the operator 
- create a mesh
- deploy a deployment or stateful set into a watched namespace with the above annotation
- the resource should be created and deleted as desired.  Check to ensure no sidecar was appended to the pod.

Im sure there is a way to abstract the deployment/stateful set switch block but im unsure how to do that.  I think we need to abstract this to a client.Object.  Would love to pair program with someone if anyone wants to refactor this.


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: docker.greymatter.io/development/gm-operator
-  newTag: 0.0.1
+  newName: docker.greymatter.io/internal/gm-operator
+  newTag: latest
```
bcmendoza, line 16: Was this triggered by running the dev script or did you change it manually? This change needs to be reverted.

hack/sample-exclude-from-mesh.yaml
```diff
@@ -0,0 +1,27 @@
...
+      containers:
+      - image: docker.greymatter.io/internal/fibonacci:latest
+        imagePullPolicy: Always
+        name: fibonacci
+        ports:
+        - containerPort: 8080
+          protocol: TCP
+        resources: {}
+      imagePullSecrets:
+        - name: gm-docker-secret
```
bcmendoza, line 27: nit: new line

bcmendoza (changes requested):
Let's talk about our UX for this. Could we use the same label for determining whether a service is exposed by edge (default), is itself an edge, unreachable (i.e. not routable by an edge), or excluded? Otherwise, if we go the route of having an `exclude-from-edge=true` label, then we'll need similar labels for each mode, and then we'll need to do some weird logic for determining what to do when multiple modes are enabled.

Here's an initial idea:
```
greymatter.io/network-mode: route // routable via edge ingress; this is always applied by default
greymatter.io/network-mode: gateway // routable via its own ingress
greymatter.io/network-mode: system // not routable from an ingress, but still included in a mesh
greymatter.io/network-mode: exclude // excluded entirely from a mesh
```

I still need to walk through the code in `workloads.go`. I agree that it's getting overly complex, and we should tidy things up there. We should pair on it.

config/manager/kustomization.yaml
moosecanswim: I think this may have been from a botched rebase.  Will change it.

moosecanswim:
@bcmendoza `greymatter.io/network-mode: gateway // routable via its own ingress` for this do mean the operator will create a new edge for it and route to this service?  Or assuming an ingress will be manually created for it?

bcmendoza:
> @bcmendoza `greymatter.io/network-mode: gateway // routable via its own ingress` for this do mean the operator will create a new edge for it and route to this service? Or assuming an ingress will be manually created for it?

Hmm… I guess implementing that would really depend on the ingress controller. We shouldn’t worry about that one for now though. Just the 3 other options.

dontlaugh:
@moosecanswim any update on this?

moosecanswim:
@dontlaugh working through yesterday.  I'm currently updating the fabric cue files so that mesh objects will be created based on the `networkMode` label.

moosecanswim:
@dontlaugh @bcmendoza Updated this PR to use `greymatter.io/network-mode` label.

- `exclude`: will not add sidecar (mesh configs not created either)
- `internal`: sidecar added and mesh configs for sidecar -> service created as normal (No catalog entry tho i think we may want it to have a tile)
- `route`: this is default and has the same behavior as if the label is not present.  Adds sidecar, greates mesh configs for edge -> sidecar -> service + catalog entry

**Testing**
similar as before start the operator and apply the file.  To test the changes change the label in the file then re create it.

dontlaugh:
@moosecanswim marking this as wip right now. We might consider translating what we learned into design. This might be post-MVP work.

PR #37, opened 2021-11-09
Simplify generating manifests for K8s

bcmendoza:
This PR fills in a missing step in the process for installing the operator in K8s by adding CFSSL as the cert signer for the certs used by the operator's webhook server. As a result, nearly complete manifests can be generated using the command `kustomize build config/k8s`. For details, see the [README](https://github.com/greymatter-io/operator/pull/37/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R7).

I also cleaned up a lot of the kustomize usage and removed unnecessary options in favor of sensible defaults.


dontlaugh (changes requested):
Please start wrapping your markdown at ~80-100 chars

README.md
```diff
@@ -4,6 +4,29 @@ Grey Matter Operator is a Kubernetes operator that enables using `greymatter.io/
...
+
+```
+kustomize build config/k8s
+```
+
+(NOTE: If deploying to OpenShift, you can run `kustomize build config/openshift`.)
+
+As a convenience, `kustomize` may be downloaded to this repo's `bin` directory by using the `make kustomize` target.
+
+When applying these manifests to your Kubernetes cluster, you'll also need to create a [docker-registry secret](https://kubernetes.io/docs/concepts/configuration/secret/#docker-config-secrets) so the Kubelet has access to pull Grey Matter images. Provided that you have the necessary Docker username and password, these can be created with the following command:
```
dontlaugh, line 19: Wrap the markdown. Something like this 

```suggestion
When applying these manifests to your Kubernetes cluster, you'll also need to create a 
[docker-registry secret](https://kubernetes.io/docs/concepts/configuration/secret/#docker-config-secrets) 
so the Kubelet has access to pull Grey Matter images. Provided that you have the 
necessary Docker username and password, these can be created with the following command:
```

config/default/kustomization.yaml
```diff
@@ -1,11 +1,8 @@
-# Adds namespace to all resources.
+# Adds namespace to all resources, overriding the 'system' namespace.
```
dontlaugh, line 1: Right now the manifest we yield is not that big:

```
% ./bin/kustomize build config/k8s | wc -l      
604
```
What are the chances we can move this into a native Go template? Does kustomize do anything else special for us?

pkg/webhooks/webhooks.go
```diff
@@ -1,21 +1,218 @@
...
+	if wl.disableCertGen {
+		wl.register()
+		return nil
+	}
+
+	// Patch the opaque secret with our previously loaded signed certs
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "gm-controller-manager-service-cert",
+			Namespace: "gm-operator",
```
dontlaugh, line 63: Any chance we'll need to make this configurable?

pkg/webhooks/webhooks.go
```diff
@@ -1,21 +1,218 @@
...
+
+func New(cl client.Client, i *installer.Installer, c *cli.CLI, get func() *webhook.Server, disableCertGen bool) (*Loader, error) {
+	wl := &Loader{Client: cl, Installer: i, CLI: c, getServer: get, disableCertGen: disableCertGen}
+
+	var err error
+	if !wl.disableCertGen {
+		err = wl.loadCerts()
+	}
+
+	return wl, err
```
dontlaugh, line 48: Use a narrower scope for the error
```suggestion
	if !wl.disableCertGen {
		if err := wl.loadCerts(); err != nil {
		    return nil, err
		}
	}

	return wl, nil
```

dontlaugh (changes requested):
Looks good overall. A few changes and questions.

pkg/webhooks/webhooks.go
bcmendoza: I don't think so. The user really shouldn't be messing with these. We may want to be more deterministic by delegating the creation of these resources to the operator itself, which should limit the amount of manifests needed for the initial install.

config/default/kustomization.yaml
bcmendoza: We need Kustomize for two reasons:
1. Operator SDK generates files that are only compatible with it. There are template markers throughout our code that have 1:1 mappings to Kustomize values. I don't feel safe quite yet about moving away from the Operator SDK's generators, mainly in terms of its Openshift nuances.
2. In general, Kustomize makes it easier to add custom overlays and patches on existing manifests. I prefer it to needing to re-invent the wheel with Go templates.

Kustomize also is pretty widely adopted, and even has its own command in `kubectl`.

With all that said, I COULD see us eventually move to Cue, but only after we feel comfortable about not relying on the Operator SDK generators.

config/default/manager_webhook_patch.yaml
```diff
@@ -14,10 +14,15 @@ spec:
...
+          name: webhook-cert
           readOnly: true
       volumes:
-      - name: cert
+      - name: webhook-cert
         secret:
           defaultMode: 420
-          secretName: webhook-server-cert
+          items:
+          - key: tls.crt
```
dontlaugh, line 24: Can we add a comment here about where these certs come from

config/k8s/manager_image_patch.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Deployment
+metadata:
+  name: gm-controller-manager
+  namespace: gm-operator
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        image: docker.greymatter.io/internal/gm-operator:latest-kops
```
dontlaugh, line 11: Can we use `latest`? It'd be better if we didn't have to use a special build here.

config/k8s/webhook_secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: gm-controller-manager-service-cert
+  namespace: gm-operator
+data:
+  tls.crt: ""
```
dontlaugh, line 7: Add a comment about how these certs work, where they come from, etc.

config/manager/bootstrap_config.yaml
```diff
@@ -1,12 +1 @@
...
-health:
-  healthProbeBindAddress: :8081
-metrics:
-  bindAddress: 127.0.0.1:8080
-webhook:
-  port: 9443
-leaderElection:
-  leaderElect: true
-  resourceName: 715805a0.greymatter.io
 imagePullSecretName: gm-docker-secret
```
dontlaugh, line 1: Add a comment to this file. Not often you see a one-line yaml file.

pkg/bootstrap/bootstrapconfig_types.go
```diff
@@ -18,21 +18,51 @@ package bootstrap
...
+	DisableWebhookCertGeneration bool `json:"disableWebhookCertGeneration"`
+	// The name of the cluster ingress (configs.openshift.io/v1.Ingress) in an Openshift environment.
+	// This is used to determine the domain of the cluster. Defaults to "cluster" if not set.
+	// This is ignored in non-Openshift environments.
 	ClusterIngressName string `json:"clusterIngressName"`
 }
+
+// Load our bootstrap config and read any manager options that have not already been set in manager.Options.
+// The values set in manager.Options will always take precedence.
+func LoadWithManagerOptions(path string, options manager.Options) (BootstrapConfig, manager.Options, error) {
```
dontlaugh, line 48: How many places is this function called? This function is a strange abstraction. An alternative to this is using a global, and getting rid of this function altogether. Something like

```go
var defaultBootstrapConfig =  BootstrapConfig{
		ControllerManagerConfigurationSpec: cfg.ControllerManagerConfigurationSpec{
			// Required as an empty config since options.AndFrom below does not check for nil
			LeaderElection: &basecfg.LeaderElectionConfiguration{},
		},
		// Set default config values
		ClusterIngressName: "cluster",
	}
```
And just invoke the rest of the function's logic directly wherever you're calling this.

pkg/webhooks/cfssl.go
```diff
@@ -0,0 +1,176 @@
...
+		Hosts:      []string{"greymatter.io"},
+		KeyRequest: &csr.KeyRequest{A: "rsa", S: 2048},
+		CA:         &csr.CAConfig{Expiry: "8760h"},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	os.Setenv("CFSSL_CA", string(caCert))
+	os.Setenv("CFSSL_CA_KEY", string(caKey))
```
dontlaugh, line 49: Do we _really_ have to set these as env vars? Or can we provide them to the config some other way.

pkg/webhooks/cfssl.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// serveCFSSL launches a CFSSL server for issuing TLS certs:
+// 1. for the operator's webhook server, to admit requests from the Kubelet
+// 2. TODO: for the operator to issue intermediate CAs for intra-mesh communication (if desired)
+// For now, serveCFSSL is local to the webhooks package. This will likely move into its own package later.
+// The returned bytes contain the CA returned from a running server, ensuring the server is initialized.
+func serveCFSSL() ([]byte, error) {
```
dontlaugh, line 28: We really need an object to hold all this configuration, and put methods on it for start/stop.

```go
type CFSSLServer struct {
  logger *L
  ca []byte
  caKey []byte
  listenerAddress string
}
```
The benefit here is that it lets us make this configurable, and also give us flexibility in how we request the CA cert.

Separate a `Start` method from a separate getter method that returns the cert material, if requested.

And write a constructor

pkg/webhooks/cfssl.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil, err
+	}
+
+	return []string{
+		resp.Result.Cert,
+		resp.Result.Key,
+	}, nil
+}
+
+type cfsslResp struct {
```
dontlaugh, line 120: If this structure models a json response outside of our control, note that here.

pkg/webhooks/webhooks.go
```diff
@@ -1,21 +1,165 @@
...
+	cert           string
+	key            string
+}
+
+func New(cl client.Client, i *installer.Installer, c *cli.CLI, get func() *webhook.Server, disableCertGen bool) (*Loader, error) {
+	wl := &Loader{Client: cl, Installer: i, CLI: c, getServer: get, disableCertGen: disableCertGen}
+
+	if !wl.disableCertGen {
+		// Initialize and launch CFSSL server. This will eventualy move out of the webhooks package,
+		// but can stay here for now since we're just using it to issue our webhook certs.
```
dontlaugh, line 42: Here is where we should do something like

```go
svr := NewCFSSLServer(...)
if err := svr.Start(); err != nil {
// ...
}
caBundle := svr.GetCABundle()  // nil could signal error? maybe return err, too
req := csr.CertificateRequest{
		CN:         "admission",
		Hosts:      []string{"gm-webhook-service.gm-operator.svc"},
		KeyRequest: &csr.KeyRequest{A: "rsa", S: 2048},
	}
cert, key, err := svr.RequestCert(&req)
```

I see your comment waving towards a future where we abstract this out. Let's do that now, and hoist the configuration to the call site, rather than hardcoding it internally. Here we still hard code, but it's in a context wehre configuration is accessible.

dontlaugh (changes requested):
Let's spend some time making the CFSSL stuff reusable. I think if you can find a good pattern here it'll have a big payoff.

pkg/webhooks/cfssl.go
bcmendoza: The only other option would be to write to files and provide the file paths.

dontlaugh: okie dokie

dontlaugh (approved):
Looks good

PR #40, opened 2021-11-23
Use SPIRE for mTLS

bcmendoza:
Note: The commit history is jumbled since I've had this branch in WIP for a while. I will squash and merge into a single commit when this is merged in.

This PR does the following:
- Adds SPIRE to the operator's initialization process. The SPIRE server is configured to use the CFSSL server provided trust chain with its own intermediate CA for issuing identities.
- Bootstraps all configured clusters and listeners with the necessary secret configuration for SPIRE-managed mTLS communication between mesh workloads.
- Edge still receives plaintext traffic and must be secured by the user's certificate authority of choice.
- NOTE: SPIRE is on by default. A flag will be implemented later on when we allow the user to specify their own mTLS (perhaps allow the user to enable mesh-wide mTLS with the same certs across all workloads).


dontlaugh:
Can we squash pre-merge?

dontlaugh:
PR for perms over in CLI required?

bcmendoza:
@alecholmez @dontlaugh 

For now, let's test this PR separately from the CLI. We can test it again when pinning the updated commit in the CLI.

Ensure `GREYMATTER_DOCKER_USERNAME` and `GREYMATTER_DOCKER_PASSWORD` has your Nexus credentials and then run:

```
kubectl apply -k config/context/kubernetes

kubectl create secret docker-registry gm-docker-secret \
  --docker-server=docker.greymatter.io \
  --docker-username=$GREYMATTER_DOCKER_USERNAME \
  --docker-password=$GREYMATER_DOCKER_PASSWORD \
  --docker-email=$GREYMATTER_DOCKER_USERNAME \
  -n gm-operator
```

dontlaugh:
@alecholmez I'm going to tag this issue and bring it into the sprint for Brian, he can work on this after the cli integration and while you test the basic functionality

bcmendoza:
@alecholmez @dontlaugh This is ready to test. Let me know if you'd like a quick code walkthrough.

pkg/cli/commands.go
```diff
@@ -0,0 +1,125 @@
...
+				logger.Info("apply", "type", kind, "key", key, "Mesh", mesh)
+			}
+		},
+		modify: func(out []byte) ([]byte, error) {
+			outStr := string(out)
+			if !strings.Contains(outStr, "200") {
+				return out, fmt.Errorf(string(out))
+			}
+			return out, nil
+		},
```
bcmendoza, line 31: When a `cmd` is `run()`, it returns `(string, error)`. (the string is either `stdout` or `stderr`) But `greymatter apply` never exits with an error, so we need to parse the result of `stdout` in order to know that the operation worked.

Whenever an operation fails, the operator requeues the request. This needs to happen sometimes since our current data model has object dependencies enforced by Control API. e.g. If a referenced upstream egress cluster does not yet exist, creating the referencing object is requeued on the loop that runs the commands.

Ideally, we wouldn't need to parse the command output, but instead the CLI would exit with an error whenever `greymatter apply` does not succeed. It makes sense to return an error in our case since apply only takes a single object at a time.

hack/egress-demo.yaml
```diff
@@ -59,7 +59,7 @@ spec:
         app: fibonacci
     spec:
       containers:
-      - image: docker.greymatter.io/internal/fibonacci:latest
+      - image: docker.greymatter.io/internal/responder:egress-demo
```
dontlaugh, line 62: This is fine for now, but we'll want to move these somewhere else by next week.

hack/egress-demo.yaml
```diff
@@ -59,7 +59,7 @@ spec:
...
         imagePullPolicy: Always
         name: fibonacci
```
dontlaugh, line 64: change name?

main.go
```diff
@@ -121,40 +118,67 @@ func main() {
...
-	// Initialize the webhooks Loader and add it as a runnable process to the operator so that
-	// it patches our webhook configurations after the operator starts.
-	wl, err := webhooks.New(c, inst, gmcli, mgr.GetWebhookServer, cfg.DisableWebhookCertGeneration)
+	// Start up our CFSSL server for issuing two certs:
+	// 1) Webhook server certs (unless disabled in the bootstrap config)
+	// 2) SPIRE's intermediate CA for issuing identities to workloads
+	cs, err := cfsslsrv.New(nil, nil)
 	if err != nil {
-		os.Exit(1)
+		logger.Error(err, "Failed to configure CFSSL server")
```
dontlaugh, line 133: We should wrap the error and let the caller log in one place instead of adding multiple logging statements to this function.

main.go
```diff
@@ -121,40 +118,67 @@ func main() {
...
 		logger.Error(err, "problem running operator")
+		return err
+	}
+
+	return nil
+}
+
+func main() {
+	if err := run(); err != nil {
 		os.Exit(1)
```
dontlaugh, line 182: we can log error here.

pkg/cli/cmd.go
```diff
@@ -13,8 +13,12 @@ type cmd struct {
 	stdin json.RawMessage
 	// Notifies the caller to requeue the cmd if it fails.
 	requeue bool
-	// A custom logger; if not used, the full result.cmdout will be logged.
-	log func(error)
+	// A custom logger; if not set, nothing is logged.
+	log func(string, error)
+	// If set, modifies the output before it is returned.
+	modify func([]byte) ([]byte, error)
```
dontlaugh, line 19: What was the motivation behind this new callback?

pkg/fabric/cue/api.cue
```diff
@@ -1,8 +1,8 @@
 // catalogservice
 
 #CatalogService: {
-  mesh_id: string
-  service_id: string
+  mesh_id: *MeshName | string
+  service_id: *ServiceName | string
```
dontlaugh, line 5: Where do `MeshName`, `ServiceName`, and `Zone` come from? Where are they defined?

pkg/fabric/cue/spire.cue
```diff
@@ -0,0 +1,88 @@
+MeshName: string
+ServiceName: string
```
dontlaugh, line 2: I see here. Any reason this isn't `#MeshName` and `#ServiceName`?

pkg/k8sapi/k8sapi.go
```diff
@@ -0,0 +1,111 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+var (
+	logger = ctrl.Log.WithName("k8sapi")
+)
+
+// Apply is a functional interface for interacting with the K8s apiserver in a consistent way.
+// Each client.Object argument must implement the necessary interfaces for Reader/Writer interfaces implemented by client.Client.
```
dontlaugh, line 19: Specify the interfaces by fully-qualified name here.

pkg/k8sapi/k8sapi.go
```diff
@@ -0,0 +1,111 @@
...
+	} else {
+		kind = gvk.Kind
+	}
+
+	// Set an owner reference on the manifest for garbage collection if the owner is deleted.
+	var ownerName string
+	if owner != nil {
+		ownerName = client.ObjectKeyFromObject(owner).Name
+		if err := controllerutil.SetOwnerReference(owner, obj, scheme); err != nil {
+			logger.Error(err, "SetOwnerReference", "Owner", ownerName, kind, client.ObjectKeyFromObject(obj))
```
dontlaugh, line 35: improve these error messages

pkg/k8sapi/k8sapi.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return "get", nil
+}
+
+// MkPatchAction returns a function that applies the patch specified when called.
+func MkPatchAction(patch func(client.Object) client.Object) func(client.Client, client.Object) (string, error) {
+	return func(c client.Client, obj client.Object) (string, error) {
+		key := client.ObjectKeyFromObject(obj)
+		if err := c.Get(context.TODO(), key, obj); err != nil {
+			return "get", err
```
dontlaugh, line 100: What is the error case here? Does the caller need the "get" string? If so, why?

dontlaugh (changes requested):
Just a few questions. Overall I'm comfortable merging this after the nitpicking is addressed.

pkg/cli/cmd.go
bcmendoza: This is needed to modify a fetched mesh config and then send the modified config back to Control API. It is used to update an existing listener with additional subject names: https://github.com/greymatter-io/operator/blob/spire/pkg/cli/commands.go#L44

pkg/k8sapi/k8sapi.go
bcmendoza: Can you elaborate?

pkg/fabric/cue/spire.cue
bcmendoza: I prefer to use a hash when defining definitions vs values. It's okay that these values show up in a `cue eval` since I'm not reading them in Go code

pkg/k8sapi/k8sapi.go
bcmendoza: Oh, you mean by full package path

pkg/k8sapi/k8sapi.go
bcmendoza: In the context of this function, any error resulting from Client.Get should be returned as an error to the caller, since failing to Get the resource means no patch can be applied to it (whether the object was not found or Get failed due to some other reason). The "get" string and other strings are returned to give the caller context as to where in the sequence of calls the failure occurred.

I defined an `ActionFunc` type that describes how these functions should be written and how they are used by Apply.

dontlaugh (approved):

PR #51, opened 2022-01-03
Prevent greymatter apply from running until Control API connects to Redis

bcmendoza:


dontlaugh (changes requested):
Create a more randomized name for this sharedrules key, and clean up afterwards

dontlaugh (approved):

PR #54, opened 2022-01-06
Secure bootstrap routes

bcmendoza:
Changes the operator's communication with Control API and Catalog to no longer route through edge. It also exposes Control API and Catalog service ports (5555 and 8080) to the cluster, but this is temporary if I can get the operator talking to their sidecars via mTLS (SPIRE).


pkg/version/cue.mod/outputs.cue
```diff
@@ -166,22 +166,59 @@ manifests: [...#ManifestGroup] & [
...
+        //   },
+        // }
+        // if _name == "gm-redis" {
+        //   {
+        //     name: "gm-redis"
+        //     protocol: "TCP"
+        //     port: 6379
+        //     targetPort: 6379
+        //   },
+        // }
```
alecholmez, line 221: Should we leave this commented code in here?

bcmendoza: Ah good catch. Meant to delete this

dontlaugh (approved):

alecholmez (approved):

PR #57, opened 2022-01-09
WIP Add sensible HTTP and Network filter defaults

bcmendoza:
- Observables filter enabled for all sidecars except edge
- Removed ability to select HTTP and network filters via annotation, in favor of sensible defaults
- Added `greymatter.io/ingress-tcp-port-name` annotation option for non-HTTP services. When specified, requests to a sidecar's `10808` port will proxy to the service over raw TCP


bcmendoza:
OBE

PR #58, opened 2022-01-10
WIP Remove Prometheus

bcmendoza:
Removes Prometheus from a Mesh's core installation. Prometheus is an optional dependency and shouldn't be installed directly by our operator.

TODO: How to support Prometheus if installed via another process.


bcmendoza:
OBE

PR #64, opened 2022-01-20
cue: add the ability to specify alt repos/images for core service downloads

alecholmez:
Signed-off-by: Alec Holmes <alecholmez@me.com>


config/base/deployment_config_patch.yaml
```diff
@@ -10,6 +10,7 @@ spec:
       - name: operator
         args:
         - "--config=bootstrap_config.yaml"
+        - "--leader-elect"
```
bcmendoza, line 13: @alecholmez This doesn't need to be specified here since it's already specified in `config/base/deployment/deployment.yaml`. This file is meant to be a patch that just adds the `bootstrap_config.yaml`. Those two args will be merged into the final result.

bcmendoza: Ah, I take it back. It seems entire fields get overwritten if defined in a patch. This can just stay here then.

config/base/deployment/deployment.yaml
```diff
@@ -22,8 +22,6 @@ spec:
       containers:
       - command:
         - /app/operator
-        args:
-        - --leader-elect
```
bcmendoza, line 26: @alecholmez I removed this flag because the operator doesn't actually read it anymore. Leader elect is turned on by default and otherwise can be overridden via the bootstrap config.

bcmendoza (changes requested):

config/base/deployment/deployment.yaml
alecholmez: Okay sweet that sounds good

config/config.go
```diff
@@ -63,33 +69,35 @@ var kubernetesCommand = cli.Command{
 			DockerUsername:               c.String("registry-username"),
 			DockerPassword:               c.String("registry-password"),
 			DisableWebhookCertGeneration: c.Bool("disable-internal-ca"),
+			SecretsList:                  strings.Split(c.String("pull-secrets"), ","),
```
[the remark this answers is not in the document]
alecholmez: Yup totally reasonable. I had a reason why I made it a generic secrets list but I've forgotten so we'll go with `ImagePullSecretsList` :smile:

api/v1alpha1/mesh_types.go
```diff
@@ -27,10 +27,19 @@ import (
...
 	ReleaseVersion string `json:"release_version"`
 
+	// A list of OCI image strings and their respective pull secret names.
+	// These are treated as overrides to the specified "release_version".
+	// +optional
+	Images map[string]string `json:"images,omitempty"`
+
+	// A map of pull secrets grouped by namespaces.
+	// +optional
+	PullSecrets map[string][]string `json:"pull_secrets,omitempty"`
```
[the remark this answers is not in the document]
alecholmez: I think I'm good with this. Is the reason to make `Images` a full on go struct that the cue interface will be easier to deal with in the inputs file? If that's the case I'm cool with this. The only reason I did a generic `map[string]string` was to handle image strings without worrying about nesting their structure in the yaml. I don't might the explicitness here, just want to make sure that's what we want.

2. Totally okay with renaming the imagePullSecrets. I'm not tied to that at all.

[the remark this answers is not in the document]
bcmendoza: @alecholmez Yeah, a Go struct will help for both CUE and for the user applying manifests. We get built in validation from those kubebuilder markers on the CRD.

pkg/cuemodule/base/inputs.cue
```diff
@@ -3,16 +3,28 @@ package base
...
 IngressSubDomain: *"" | string
 
-InstallNamespace: string
-WatchNamespaces: [...string]
+InstallNamespace: mesh.spec.install_namespace
+WatchNamespaces: mesh.spec.watch_namespaces
+
+// Secrets that can live in other namespaces so users aren't tied to
+// docker.greymatter.io
+ImagePullSecrets: mesh.spec.image_pull_secrets
```
bcmendoza, line 26: This isn't being read anywhere. You'll want to make sure each listed image pull secret is also added to this list: https://github.com/greymatter-io/operator/blob/main/pkg/cuemodule/base/outputs.cue#L54-L56

bcmendoza (changes requested):

api/v1alpha1/mesh_types.go
```diff
@@ -47,6 +56,21 @@ type MeshSpec struct {
...
 
+type UserToken struct {
+	Label  string              `json:"label"`
+	Values map[string][]string `json:"values"`
+}
+
+type Images struct {
+	Proxy       string `json:"proxy,omitempty"`
+	Catalog     string `json:"catalog,omitempty"`
+	Control     string `json:"control,omitempty"`
```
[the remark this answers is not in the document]
alecholmez: Oh oops, I chose not to add control-api because I figured we could derive the version from control since they should always match. But i'll add back control-api

pkg/cuemodule/base/inputs.cue
alecholmez: Oh okay cool, i'll fix this

bcmendoza (approved):
LGTM! Thanks for all your work on this. Can you clean up the commit history and ping me when it's ready?

PR #65, opened 2022-01-20
WIP K8s manifests to cue

bcmendoza:


pkg/cuemodule/meshconfigs/test_inputs.cue
```diff
@@ -0,0 +1,85 @@
+package meshconfigs
```
alecholmez, line 1: Everything is commented out in this file do we still need it for tests?

alecholmez (approved):

pkg/cuemodule/meshconfigs/test_inputs.cue
bcmendoza: Nope! That's basically a scratch file to use when evaluating with Cue's CLI.

PR #68, opened 2022-01-26
Added nix shell and updated readme

jmcudd:
This lets you spin up k3d and installs GM Operator automatically. Just run `nix-shell`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,15 +21,10 @@ To get the latest development version of the operator up and running in your Kub
 the following:
 
 ```
-
+kubectl create namespace gm-operator
+docker login docker.greymatter.io
+kubectl create secret generic gm-docker-secret --from-file=.dockerconfigjson=$HOME/.docker/config.json --type=kubernetes.io/dockerconfigjson -n gm-operator
 kubectl apply -k config/context/kubernetes
-
-kubectl create secret docker-registry gm-docker-secret \
-  --docker-server=docker.greymatter.io \
-  --docker-username=$GREYMATTER_REGISTRY_USERNAME \
-  --docker-password=$GREYMATTER_REGISTRY_PASSWORD \
-  --docker-email=$GREYMATTER_REGISTRY_USERNAME \
-  -n gm-operator
 ```
 
 The operator will be running in a pod in the `gm-operator` namespace.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -21,15 +21,10 @@ To get the latest development version of the operator up and running in your Kub
 the following:
 
 ```
-
+kubectl create namespace gm-operator
+docker login docker.greymatter.io
+kubectl create secret generic gm-docker-secret --from-file=.dockerconfigjson=$HOME/.docker/config.json --type=kubernetes.io/dockerconfigjson -n gm-operator
```
bcmendoza, line 26: We expect the environment variables `GREYMATTER_REGISTRY_USERNAME` and `GREYMATTER_REGISTRY_PASSWORD ` to be populated. It's used in other places in the project. Also, this won't work for OSX users since the `config.json` relies on the `credsStore` field (OSX keychain).

README.md
```diff
@@ -75,8 +70,7 @@ CLI](https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/).
 Go dependencies can be added with `GO111MODULE=on go mod vendor`. The Go dependencies will be
 downloaded to the gitignored `vendor` directory.
 
-Cue dependencies should be added inside of the `/pkg/version/cue.mod` directory by running `cue get
-go k8s.io/api/...` inside of that directory. The Cue dependencies will be downloaded to the
+Cue dependencies should be added inside of the `/pkg/version/cue.mod` directory by running `cue get go k8s.io/api/...` inside of that directory. The Cue dependencies will be downloaded to the
```
bcmendoza, line 73: This needs to be separate lines, as we do line breaks at 100 chars. It makes it inconvenient in this case, but it's what I've been told.

README.md
jmcudd: Oh then we will definitely want to test the nix shell on mac because it leverages this same method of creating secrets.

alecholmez:
Is there a way we can have this default to podman and fall back onto docker?

jmcudd:
> Is there a way we can have this default to podman and fall back onto docker?

So the nix shell doesn't install docker or podman due to the complexities of configuring the hypervisor on the host machine. There is nothing preventing the use of podman. If the user has an `alias docker=podman` set, then it will use podman, otherwise it will use docker.

jmcudd:
Depending on the usecase for this nix shell some other niceties would be to include some of the build tool dependencies. Currently I think it is most useful for spinning up a local k8s environment with GM installed so target applications can be installed and tested with the mesh.

README.md
jmcudd: This is resolved. I reverted the changes to keep it at 100 chars.

dontlaugh: ~80-100, technically

bcmendoza:
@jmcudd The scripts used to bootstrap the K3d environment were originally meant for local development of the operator itself. Could you define a new context in the config directory that overlays the operator deployment with a stable image (i.e. development/gm-operator:latest)? And then your Nix script could apply that context to the cluster instead of context/kubernetes.

dontlaugh (commented):
lgtm

bcmendoza:
@jmcudd LGTM as well, but could you squash the commit history into a single message? Something like "Added nix-shell script for local quickstart environment".

bcmendoza (approved):
LGTM, please use the "squash and merge" strategy when merging into main so that there is only one commit with a succinct message.

PR #69, opened 2022-01-28
Fix Cue definition validation

bcmendoza:


main.go
```diff
@@ -106,24 +106,6 @@ func run() error {
...
-	if err != nil {
-		return fmt.Errorf("failed to create initial client: %w", err)
-	}
-
-	// Initialize operator with configured options
-	mgr, err := ctrl.NewManager(restConfig, options)
-	if err != nil {
-		return fmt.Errorf("failed to initialize controller-manager: %w", err)
-	}
-
```
bcmendoza, line 126: Note: I moved `ctrl.GetConfigOrDie` further down in our main function in order to be able to run `cli.New` outside of a K8s cluster. This allowed me to run `docker run docker.greymatter.io/internal/gm-operator:latest` locally in order to debug meshconfigs initialization (i.e. loading of the CUE meshconfigs package).

config/base/deployment/deployment.yaml
```diff
@@ -21,11 +21,11 @@ spec:
       - name: gm-docker-secret
       containers:
       - command:
-        - /operator
+        - /app/operator
         args:
         - --leader-elect
         image: docker.greymatter.io/internal/gm-operator:latest
```
dontlaugh, line 27: shouldn't this be templatable

bcmendoza: I don't think so. The Dockerfile sets the working directory to `/app` and puts the operator and all artifacts in that dir. If the user wants to edit it, they can pipe to a file and manually edit.

pkg/cuemodule/load.go
```diff
@@ -12,12 +13,13 @@ import (
...
-// Initialize the path to our Cue module directory (i.e. this directory).
+// Initialize the path to our Cue module directory.
+// We expect the working directory in the container to be `/app`.
+// If not running from `/app`, this is in a unit test which needs the runtime file path.
 func init() {
-	_, filename, _, ok := runtime.Caller(0)
-	if !ok {
-		panic(fmt.Errorf("failed to retrieve path to Cue module"))
-	} else {
+	dirPath, _ = os.Getwd()
```
dontlaugh, line 20: Let's unwind this use of globals. Somehow the person standing up the operator needs to provide this path as configuration.

Ideally this config parsing/loading happens early in program startup, then gets passed down to things that need it.

You initialize some objects early in program startup, and this path to our CUE module would be appropriate to set as a field on those structs, imo.

dontlaugh (changes requested):
Let's refactor to avoid this use of globals.

config/base/deployment/deployment.yaml
dontlaugh: I won't block this PR, but if it's an image tag, it should probably be configurable.

pkg/cuemodule/load.go
```diff
@@ -10,21 +11,29 @@ import (
...
-	} else {
-		dirPath = path.Dir(filename)
-	}
-}
+// Loader loads a package from our CUE module.
+type Loader func(string) (cue.Value, error)
 
 // Loads a package from our Cue module.
 // Packages are added to subdirectories and declared with the same name as the subdirectory.
 func LoadPackage(pkgName string) (cue.Value, error) {
```
dontlaugh, line 19: Use Go comment style for exported members. `//`, space, member name.

```
// LoadPackage ...
```

dontlaugh (changes requested):
couple nitpicks

config/base/deployment/deployment.yaml
bcmendoza: OH. The image tag. That IS templatable via CLI. This is just a base and can be overridden.

dontlaugh: Thanks for clarifying :+1:

dontlaugh (approved):

PR #76, opened 2022-02-09
Bump spire version to fix Spire attestation

danielpcox:
Updated the versions of Spire images (spire-server, spire-agent, and the k8s-workload-registrar) that the operator installs to 1.2.0. This fixes an issue on some platforms where Spire would misidentify the process as non-k8s, and fail to issue an identity, causing a chain of failures when the static config of the control-api, catalog, and gm-redis sidecars fail to make connections.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+## Fixed
+
+- Updated versions of Spire images (spire-server, spire-agent, and the k8s-workload-registrar) that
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+## Fixed
```
dontlaugh, line 5: Nit: This is a `Changed`, not a fix.

dontlaugh (changes requested):
One nit. Thanks for figuring this out.

dontlaugh (approved):

PR #77, opened 2022-02-09
Increasing limits and timeouts for Rancher Desktop

danielpcox:
For whatever reason, the operator uses >=177MiB for a simple Grey Matter deployment, and then rests at ~154MiB, so I increased the memory limit to 300. The liveness and readiness probes were also causing trouble, so I increased both of their initialDelaySeconds to 120 to accommodate slower/emulated machines (such as my M1 Mac).

With these changes, the mesh comes up in ~2 minutes from scratch, using either the dockerd or containerd runtime for Rancher Desktop.

### DIFF
--- a/config/base/deployment/deployment.yaml
+++ b/config/base/deployment/deployment.yaml
@@ -31,20 +31,20 @@ spec:
           httpGet:
             path: /healthz
             port: 8081
-          initialDelaySeconds: 15
+          initialDelaySeconds: 120
           periodSeconds: 20
         readinessProbe:
           httpGet:
             path: /readyz
             port: 8081
-          initialDelaySeconds: 5
+          initialDelaySeconds: 120
           periodSeconds: 10
         resources:
           limits:
             cpu: 200m
-            memory: 100Mi
+            memory: 300Mi
           requests:
             cpu: 100m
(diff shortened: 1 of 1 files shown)


dontlaugh (commented):
Update Changeling.

This is generous, but seems okay for the operator.

I'd hesitate before putting values this long for our other daemons.

config/base/deployment/deployment.yaml
```diff
@@ -31,20 +31,20 @@ spec:
...
             port: 8081
-          initialDelaySeconds: 15
+          initialDelaySeconds: 120
           periodSeconds: 20
         readinessProbe:
           httpGet:
             path: /readyz
             port: 8081
-          initialDelaySeconds: 5
+          initialDelaySeconds: 120
```
dontlaugh, line 40: I wonder if the operator even needs a readiness probe at all? Nothing is waiting for it afaik, and it will continuously attempt to reconcile it's state. 🤔

bcmendoza: It might be needed by the Kubelet which hits the operator’s webhook server.

dontlaugh (approved):

PR #79, opened 2022-02-10
swapping out deprecated flag in spire server health check

danielpcox:
`--registrationUDSPath` is a deprecated way of specifying the UDS path, and we recently updated the Spire components. The new flag is `-socketPath`. This fixes a bug where the liveness probe on spire server would eventually fail, causing Kubernetes to kill the container.

### DIFF
--- a/pkg/installer/spire.yaml
+++ b/pkg/installer/spire.yaml
@@ -70,7 +70,7 @@ statefulset:
               command:
               - /opt/spire/bin/spire-server
               - healthcheck
-              - --registrationUDSPath=/run/spire/socket/registration.sock
+              - -socketPath=/run/spire/socket/registration.sock
             failureThreshold: 2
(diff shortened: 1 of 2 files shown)


pkg/installer/spire.yaml
```diff
@@ -70,7 +70,7 @@ statefulset:
               command:
               - /opt/spire/bin/spire-server
               - healthcheck
-              - --registrationUDSPath=/run/spire/socket/registration.sock
+              - -socketPath=/run/spire/socket/registration.sock
```
tilleryd, line 73: does this work without the double dash?

danielpcox: Yes

danielpcox: Also I tested it, 'cause I was wondering myself

PR #82, opened 2022-02-11
Refactor main.go

dontlaugh:
- Move defaultBootstrapConfig to function scope
- Move func main to the top, separate out flags

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,8 +62,8 @@ func init() {
 
 // Global config flags
 var (
-	configFile  string
-	development bool
+	configPath string
+	zapDevMode bool
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


dontlaugh:
We'll use rebase to keep these commits separate

main.go
```diff
@@ -73,18 +60,32 @@ func init() {
...
+		os.Exit(1)
+	}
+}
+
 func run() error {
-	flag.StringVar(&configFile, "config", "", "The operator will load its initial configuration from this file if defined.")
-	flag.BoolVar(&development, "development", false, "Run in development mode.")
+	flag.Parse()
+
+	flag.StringVar(&configPath, "configPath", "", "The operator will load its initial configuration from this file if defined.")
```
bcmendoza, line 79: You’ll need to update the manifest as well.

bcmendoza (changes requested):

main.go
dontlaugh: Alec asked me to run `make generate manifests`. It yielded no diff.

alecholmez (approved):
Nice this looks good

PR #84, opened 2022-02-14
Remove unix WorkloadAttestor from spire config

dontlaugh:
Since our spire config is all-k8s, we shouldn't need this workload
attestor. This might save us some spam in the logs.


bcmendoza (approved):
LGTM, confirmed this removed the noise in the agent containers.

PR #89, opened 2022-02-17
Add annotation for disabling sidecar configuration

danielpcox:
This was necessary to make clear instructions for GitOps-based deployments in our docs, and trivial enough to do in passing.

We've added this annotation to the operator for Deployments and StatefulSets:

```
metadata:
  annotation:
    greymatter.io/configure-sidecar: "false"
```

If set to "false", the operator will inject a sidecar but leave it unconfigured so users can manually configure it.

Also in this PR

- Changelog update
- Some refactoring of magic strings into constants

### DIFF
--- a/pkg/wellknown/wellknown.go
+++ b/pkg/wellknown/wellknown.go
@@ -1,0 +1,8 @@
+package wellknown
+
+const (
+	ANNOTATION_CONFIGURE_SIDECAR = "greymatter.io/configure-sidecar"
+	ANNOTATION_LAST_APPLIED      = "greymatter.io/last-applied"
+	LABEL_CLUSTER                = "greymatter.io/cluster"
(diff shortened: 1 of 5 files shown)


pkg/wellknown/wellknown.go
```diff
@@ -0,0 +1,8 @@
+package wellknown
+
+const (
```
dontlaugh, line 3: Is this generated? If so, is there any way we can add a comment to this effect?

dontlaugh (approved):
one non-blocking comment

pkg/wellknown/wellknown.go
danielpcox: Not generated. Might be worth looking into because there's some duplication between the go package and the CUE module.

dontlaugh (approved):

PR #92, opened 2022-02-23
replace `kustomize build` with `kubectl kustomize`

danielpcox:
[sc-10466]

"Hello Dilbert, I'm the vice-president in charge of immediately implementing ideas..."  
--Dilbert Episode 2

Here's a PR. Do not trust this PR - Make sure this is what you actually want. In particular, please confirm that the CI build container has a new enough version of kubectl that it can still build the operator after this PR.

I'm just throwing something up there because when I went through all of the Operator PRs, two of them stood out as handy and completely trivial, so I resolved to just do them right now. I just naively applied what Brian said here:
https://app.shortcut.com/grey-matter/story/10466/operator-replace-kustomize-cli-with-kubectl

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,12 +44,8 @@ kubectl apply -k config/context/kubernetes --dry-run=client -o yaml
 (NOTE: If deploying to OpenShift, you can replace `config/context/kubernetes` with
 `config/context/openshift`.)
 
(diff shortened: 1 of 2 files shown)


bcmendoza (approved):
LGTM! Much more convenient to use `kubectl kustomize` vs `kustomize build`.

PR #93, opened 2022-03-07
Use ubuntu-2110-ci buildkite container

dontlaugh:
2104 was built from a non-LTS release, so we upgrade it with the next non-LTS release.

Hopefully we can get rid of non-LTS releases entirely with the next LTS (22.04)


dontlaugh:
@danielpcox take a look at this, just so you're aware of how it works

PR #94, opened 2022-03-08
Submodules cannot pin to a commit, only track HEAD

dontlaugh:
We needed to create a branch pre_refactor in the greymatter-cue repo, and we use
that as our submodule. HEAD should be this commit

    36dd308eefd7585fafa6139ce2c29755e674fd58

Once our CUE refactor is done, we can tag releases in greymatter-cue
which submodules can also use.


dontlaugh:
@danielpcox This fixes a regression introduced here https://github.com/greymatter-io/operator/pull/93

PR #97, opened 2022-04-05
CUE refactor

danielpcox:
Most of the CUE refactor, though not yet with all of the features restored that I broke during the process (detailed here: https://app.shortcut.com/grey-matter/story/11585/resurrect-skipped-operator-features )

## Changed

- Completely new CUE for both Kubernetes manifests and Grey Matter config, currently still rooted
  under `pkg/cuemodule/new_structure` (and will eventually be moved into the root once the Spire
  support is complete) following a philosophy of readability and malleability.
- Operator Go now loads CUE generically, such that nearly all changes to deployment configuration
  require no changes to the Go.
- Deployment assist is now opt-in, with labels `"greymatter.io/inject-sidecar-to": "<port num>"`
  (default is to omit the label) and `"greymatter.io/configure-sidecar": "<boolean>"` (default
  "true".)
- Operator itself now runs as a StatefulSet rather than a Deployment.
- One-Mesh/One-Operator - Multiple meshes require multiple operators (they do yet not play nicely
  together, but the target for multi-mesh support is to support graceful interop.)
- For now, automatically deploys the default Mesh described in the CUE source shortly after startup.
  Later, this will be an opt-in flag in the CUE itself so users can decide.
- Switch to ECDSA 256 certificates for speed and security.

## To Test

Follow the updated "Quick Install" block of the README.


main.go
```diff
@@ -87,6 +89,10 @@ func run() error {
 	// We have to call Parse late for some reason
 	flag.Parse()
 
+	// Immediately load all CUE
+	operatorCUE, initialMesh := cuemodule.LoadAll(cueRoot) // panics if unsuccessful
```
dontlaugh, line 93: Let's not panic. Return three params, and log fatal.

pkg/cli/cli.go
```diff
@@ -23,14 +24,14 @@ var (
 // CLI exposes methods for configuring clients that execute greymatter CLI commands.
 type CLI struct {
 	*sync.RWMutex
-	clients         map[string]*client
-	meshconfigsTmpl cue.Value
-	mTLSEnabled     bool
+	client      *Client
+	operatorCUE cuemodule.OperatorCUE
```
dontlaugh, line 28:
```suggestion
	operatorCUE *cuemodule.OperatorCUE
```
This feels like it should be a pointer. Passing a copied value could trip us up.

pkg/cli/cli.go
```diff
@@ -23,14 +24,14 @@ var (
...
+	mTLSEnabled bool
 }
 
 // New returns a new *CLI instance.
 // It receives a context for cleaning up goroutines started by the *CLI.
-func New(ctx context.Context, load cuemodule.Loader, mTLSEnabled bool) (*CLI, error) {
+func New(ctx context.Context, operatorCUE cuemodule.OperatorCUE, mTLSEnabled bool) (*CLI, error) {
```
dontlaugh, line 34: Cool :+1:

pkg/cli/cli.go
```diff
@@ -97,172 +92,89 @@ func (c *CLI) configureMeshClient(mesh *v1alpha1.Mesh, flags ...string) error {
...
-// along with all ingress and egress cluster routes derived from the given annotations and containers.
-func (c *CLI) RemoveService(mesh, workload string, obj runtime.Object) {
-	c.RLock()
-	defer c.RUnlock()
-
-	cl, ok := c.clients[mesh]
-	if !ok {
-		logger.Error(fmt.Errorf("unknown mesh"), "failed to remove fabric objects for workload", "Mesh", mesh, "Workload", workload)
+	// we also skip configuration if we're explicitly told to
+	configureSidecar := annotations[wellknown.ANNOTATION_CONFIGURE_SIDECAR]
```
dontlaugh, line 169: Whatever's handing us these annotations has an opportunity to be polite:

* scan the greymatter.io/operator prefix (or whatever it is)
* WARN on bad values, e.g. "F4Lse"

pkg/installer/install.go
```diff
@@ -25,138 +19,86 @@ func (i *Installer) ApplyMesh(prev, mesh *v1alpha1.Mesh) {
...
-					delete(i.namespaces, namespace)
-					// TODO: Remove the Docker image pull secret when a watch namespace is removed.
-				}
-			}
+		k8sapi.Apply(i.k8sClient, secret, mesh, k8sapi.GetOrCreate)
+		// TODO reverse-Chesterton's fence: I don't understand why this _wasn't_ done in the old operator
+		for _, watched_ns := range mesh.Spec.WatchNamespaces {
+			secret := i.imagePullSecret.DeepCopy()
+			secret.Namespace = watched_ns
+			logger.Info("DEBUG:", "secret", secret)
```
dontlaugh, line 31: This is just the secret's name, right? not the value? let's not log the value.

pkg/installer/installer.go
```diff
@@ -32,11 +27,10 @@ var (
 
 // Installer stores a map of version.Version and a distinct version.Sidecar for each mesh.
 type Installer struct {
-	*sync.RWMutex
-	*cli.CLI
-	client client.Client
+	*cli.CLI  // Grey Matter CLI
+	k8sClient client.Client
```
dontlaugh, line 31: Let's make this k8sClient a pointer

dontlaugh (changes requested):
An improvement, overall. I have a few nitpicks but after that I'd be good to merge.

----

For the future, ff you find a way to do the following next, it would be good:

* `panic` in the CUE processing stuff, if I understand correctly, is fine, since evaluating CUE is deterministic, and we can be assured that "it would never happen". But it would be good to find a way to return more errors in there, and log at a higher level.
* The `Cmd` and apply abstractions seem to introduce more async stuff than we need. But I could be wrong.

dontlaugh (changes requested):
oh, and squash

config/context/local-refactored/patches.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: StatefulSet
+metadata:
+  name: gm-operator
+  namespace: gm-operator
+spec:
+  template:
+    spec:
+      containers:
+      - name: operator
+        image: "docker.greymatter.io/internal/gm-operator:local_refactored"
```
alecholmez, line 11: Do we want this checked in? We might want to make a note to clean this up

pkg/cli/cli.go
```diff
@@ -97,172 +92,89 @@ func (c *CLI) configureMeshClient(mesh *v1alpha1.Mesh, flags ...string) error {
 	defer c.Unlock()
 
 	// Close an existing cmds channel if updating
```
alecholmez, line 94: this comment might be out of date?

pkg/cli/cli.go
```diff
@@ -97,172 +92,89 @@ func (c *CLI) configureMeshClient(mesh *v1alpha1.Mesh, flags ...string) error {
...
-func (c *CLI) ConfigureService(mesh, workload string, obj runtime.Object) {
-	c.RLock()
-	defer c.RUnlock()
-
-	cl, ok := c.clients[mesh]
-	if !ok {
-		logger.Error(fmt.Errorf("unknown mesh"), "failed to configure fabric objects for workload", "Mesh", mesh, "Workload", workload)
+func (c *CLI) ConfigureSidecar(operatorCUE cuemodule.OperatorCUE, name string, metadata metav1.ObjectMeta) {
+	annotations := metadata.Annotations
+	logger.Info("DEBUG: Attempting to configure sidecar with values", "name", name, "annotations", annotations)
```
alecholmez, line 123: Maybe change this to `logger.Debug`?

pkg/cli/cli.go
```diff
@@ -97,172 +92,89 @@ func (c *CLI) configureMeshClient(mesh *v1alpha1.Mesh, flags ...string) error {
...
-	if c.mTLSEnabled {
-		for _, listener := range objects.LocalEgresses {
-			cl.controlCmds <- mkInjectSVID(mesh, fmt.Sprintf("%s.%s", mesh, workload), listener)
+// UnconfigureSidecar removes fabric objects, disconnecting the workload from the mesh specified
+func (c *CLI) UnconfigureSidecar(operatorCUE cuemodule.OperatorCUE, name string, metadata metav1.ObjectMeta) {
+	annotations := metadata.Annotations
+	logger.Info("Unconfiguring sidecar with values", "name", name, "annotations", annotations)
+	injectedSidecarPortString, injectSidecar := annotations[wellknown.ANNOTATION_INJECT_SIDECAR_TO_PORT]
+	var injectedSidecarPort int
+	if injectSidecar {
```
alecholmez, line 157: This block looks like it could be a function, I think you use it above as well?

pkg/cli/client.go
```diff
@@ -3,44 +3,55 @@ package cli
...
-		f:           f,
 	}
 
-	// Consume commands to send to Control
-	go func(ctx context.Context, controlCmds chan cmd) {
+	// Apply core Grey Matter components from CUE
+	// This just dumps them on the channel, so it will block until the consumer is ready
+	go func() {
+		// by this point, GM has already been unified with THE mesh this operator manages
+		// TODO Confirm the one-operator <=> one-mesh concept, then you can delete these comments:
```
alecholmez, line 39: These still valid?

pkg/cli/client.go
```diff
@@ -54,12 +65,12 @@ func newClient(tmpl cue.Value, mesh *v1alpha1.Mesh, flags ...string) (*client, e
 			case <-ctx.Done():
 				return
 			default:
-				if _, err := (cmd{
+				if _, err := (Cmd{
 					// Create a NOOP shared_rules object to ensure that we can write to Control.
```
alecholmez, line 69: I'd maybe move this comment above the `if _, err := (Cmd{` it helps with readability

pkg/cli/client_test.go
```diff
@@ -1,77 +1,67 @@
...
-
-	c.ConfigureService("mesh", "mock", deployment)
-	c.RemoveService("mesh", "mock", deployment)
-
-	c.RemoveMeshClient("mesh")
-
-	time.Sleep(time.Second * 5)
-	cancel()
-}
+//func TestNewClient(t *testing.T) {
```
alecholmez, line 7: Maybe delete if we don't need this? or this is in limbo until spire is done

pkg/cuemodule/load.go
```diff
@@ -1,16 +1,292 @@
...
 
+type OperatorCUE struct {
+	// cue.Value for all of k8s/outputs containing k8s manifest objects
+	K8s cue.Value
+
+	// cue.Value for all of gm/outputs containing Grey Matter config objects
+	GM cue.Value
+}
+
+func LoadAll(cuemoduleRoot string) (OperatorCUE, *v1alpha1.Mesh) {
```
alecholmez, line 32: Nice, maybe add some go doc strings above these functions so they look nice in the `godoc`

pkg/installer/install.go
alecholmez: This is probably worthy of at least a `logger.Debug()`, maybe keep it out of the info level if its just the name

alecholmez (changes requested):
Nothing major from me looks great. 

Just maybe one thing, the `logger.Info("DEBUG: ")` statements can become `logger.Debug()` so they are hidden

config/context/local-refactored/patches.yaml
danielpcox: For this PR only, yes I want to check it in. I will clean it up later.

main.go
```diff
@@ -162,13 +162,13 @@ func run() error {
 	}
 
 	// Initialize manifests installer.
-	inst, err := installer.New(c, operatorCUE, initialMesh, cueRoot, gmcli, cfssl, cfg.ClusterIngressName)
+	inst, err := installer.New(&c, operatorCUE, initialMesh, cueRoot, gmcli, cfssl, cfg.ClusterIngressName)
```
dontlaugh, line 165: jack_nicholson_nodding.gif

pkg/cli/cli.go
danielpcox: I don't think so. I just changed it from dealing with potentially many meshes to just one, but the context still needs cancelling for that one mesh.

pkg/cli/cli.go
dontlaugh: for now, let's leave this alone unless Daniel wants to change it

dontlaugh (commented):
Squash this down, force push, and I'll approve

dontlaugh (dismissed):

dontlaugh:
oh I didn't see the failing build. let's try to get that green, even if it takes some time here.

PR #103, opened 2022-04-21
Retry mesh apply periodically until it succeeds

danielpcox:
Also fix the ci build script so it runs integration tests on the image it just built.


pkg/k8sapi/k8sapi.go
```diff
@@ -51,14 +51,15 @@ func Apply(c *client.Client, obj, owner client.Object, action ActionFunc) {
 		} else {
 			logger.Error(err, act, kind, client.ObjectKeyFromObject(obj))
 		}
-		return
+		return err
```
dontlaugh, line 54: If we want to log differently here, we can use a closure (or a standalone helper func), to return a default in the case of empty string.

You can also wrap errors to make tracking them down a little easier later.
```go
strOrElse := func(str, orElse string) {
    if name == "" {
      return orElse    
    }
    return str
}
logger.Error(err, act, "Owner", strOrElse(ownerName, "MISSING"), kind, client.ObjectKeyFromObject(obj))
return fmt.Errorf("action failed: %w", err)
```

dontlaugh (dismissed):

PR #105, opened 2022-04-26
ConfigMap that can be used to override config

danielpcox:
As per this conversation: https://greymatter.slack.com/archives/C07TJSLQ0/p1651009645580309?thread_ts=1650988828.618659&cid=C07TJSLQ0

I've made a new ConfigMap that gets mounted into the operator at `/app/overrides.cue`, allowing a user to override values. At the moment, the point is to let the integration tests flip switches _somehow_, and may not end up being the ultimate solution.

~Personally, I'm not very happy with this, but it'll do for now.~ I've found a version of this I'm happy enough with. See below.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,4 +35,4 @@ COPY --from=builder /workspace/pkg/cuemodule/k8s /app/k8s
 COPY --from=builder /workspace/greymatter /bin/greymatter
 USER 1000:1000
 
-ENTRYPOINT ["/app/operator"]
+CMD ["/app/operator"]
--- a/pkg/cuemodule/inputs.cue
+++ b/pkg/cuemodule/inputs.cue
@@ -10,8 +10,8 @@ import (
 
 config: {
   // Flags
-  spire: bool | *true // enable Spire-based mTLS DEBUG - the default should be false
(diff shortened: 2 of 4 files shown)


danielpcox:
I've just seen a simpler way to do this that's not much more work. Standby...

danielpcox:
Now you can just do this, which is what I was hoping would work all along, because these tags now get propagated into a file on disk inside the operator container:
```
(
cd pkg/cuemodule
cue eval -c ./k8s/outputs --out text -t operator_image=danielpcox/gm-operator:latest -t spire=false -t auto_apply_mesh=false -e operator_manifests_yaml | kubectl apply -f -

kubectl create secret docker-registry gm-docker-secret \
  --docker-server=docker.greymatter.io \
  --docker-username=$GREYMATTER_REGISTRY_USERNAME \
  --docker-password=$GREYMATTER_REGISTRY_PASSWORD \
  --docker-email=$GREYMATTER_REGISTRY_USERNAME \
  -n gm-operator
)
```

And as a bonus, you can override this ConfigMap manually later with whatever you want.

danielpcox:
@jack-bischoff

jack-bischoff (approved):

PR #106, opened 2022-04-27
Operator graceful restart

danielpcox:
Solves [sc-13114] by reloading a deployed Mesh CR if there is one to restore its state, and not auto-applying a mesh if this has happened.

Also incidentally completed [sc-13231] by reloading the OperatorCUE before attempting to unify it with the updated mesh. This avoids a conflict between the concrete values of the old and new versions of the Mesh.


danielpcox:
Left to do is reloading the SidecarList for populating Spire subjects.

1. Store the SidecarList in the Mesh whenever it changes
2. Reload it when we reload the Mesh on restart

pkg/mesh_install/installer.go
```diff
@@ -117,9 +117,27 @@ func (i *Installer) Start(ctx context.Context) error {
...
+	meshList := &v1alpha1.MeshList{}
+	if err := (*i.k8sClient).List(context.TODO(), meshList); err != nil {
+		logger.Error(err, "failed to list all meshes for state restoration - check operator permissions")
+	}
+	for _, mesh := range meshList.Items {
+		if mesh.Name == i.Mesh.Name {
+			logger.Info("Mesh already deployed. Reloading values.", "Name", mesh.Name)
+			i.Mesh = &mesh // load the live version of the mesh
+			// immediately update OperatorCUE and the SidecarList
+			i.OperatorCUE.UnifyWithMesh(i.Mesh)
```
dontlaugh, line 131: This method should return an error

danielpcox:
This is ready for review. I've tested it myself by restarting the operator and confirming the values were correctly loaded, as well as deploying/undeploying a workload to confirm that sidecar injection functions correctly both before and after an operator restart.

To test it yourself, follow the Quick Install steps to launch the operator, and notice that restarting the operator restores it to the same state it was in before the restart, eventually.

dontlaugh (changes requested):
Update the CHANGELOG

dontlaugh (approved):
If this passes test, merge it

PR #107, opened 2022-04-28
Optimize and expand integration tests

jack-bischoff:
Please see https://buildkite.com/greymatter/operator/builds/554 for the status of the tests.

This pr adds a few features. 
First, the wait times for async actions, like
installing the operator, are minimized by replacing estimated sleeps with
polling. Particularly, the wait_for_resource and curl_retry use polling with a
few second delay to effectively wait for pods and other resources to become
available and for servers to begin accepting requests, respectively.

Second, the pr adds stricter and more test assertions. These include actions like
testing for existence (and lack thereof) of pods, testing network
accessibility, and testing for readiness by a timeout (failing to be ready by a
timeout likely indicates an error as the timeouts are generous). 

Lastly, the pr expands the testing profile and facilitates future expansions by generating n pipelines and creating n k3s
clusters, where n is the number of operator tests or configurations which need
a clean machine. It does so by tagging the ec2 with the test name and including
that tag in the pipeline; an extension of the same pattern used to generate and
link just the one ec2.

There are currently three separate integration tests: one is a default installation, the other is a no spire installation, and the last creates and deletes mesh CRs to if creating meshes deploy the gm services properly and if deleting them destroys services properly.

Note, one test assertion was disabled for the tear down scenario due to a bug that Daniel may be working on. Once that fix gets merged in, then that test can be re-enabled.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -1,9 +1,16 @@
 steps:
-  - label: ":shipit: build and release"
+  - label: "run unit tests"
     agents:
       container_image: ubuntu-2110-ci
     commands: 
+      - ./scripts/cibuild bootstrap
       - ./scripts/cibuild test
+  - wait
+  - label: ":shipit: build and release"
+    agents:
+      container_image: ubuntu-2110-ci
+    commands:
+      - ./scripts/cibuild bootstrap
       - ./scripts/cibuild build
       - ./scripts/cibuild export_container "${BUILDKITE_PIPELINE_SLUG}_${BUILDKITE_BUILD_NUMBER}.tar" "docker.greymatter.io/development/gm-operator:latest"
       - ./scripts/cibuild release
@@ -12,8 +19,6 @@ steps:
         - exit_status: "*"
           limit: 2
   - wait
-  - label: "launch ec2 and run integration tests"
+  - label: "run integration tests"
     commands:
-      - ./scripts/cibuild launch_k3s
-      - ./scripts/cibuild generate_integration_pipeline | buildkite-agent pipeline upload
-
+      - ./scripts/cibuild generate_integration_tests
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -2,10 +2,8 @@
 
 set -e
 
-cmd_test () {
+cmd_bootstrap () {
   go mod vendor
-  ./scripts/test lint
-
   (
     cd ./pkg/cuemodule/cue.mod
     cue get go github.com/greymatter-io/operator/api/...
@@ -13,9 +11,15 @@ cmd_test () {
   )
 
   git submodule update --init --recursive --remote
+}
+
+# Runs go tests
+cmd_test () {
+  ./scripts/test lint
   ./scripts/test
 }
 
+# Polls the ec2 for k3s ready state
 cmd_wait_for_k3s () {
   set +e #allow errors temporarily
   echo "Waiting for k3s to become available"
@@ -32,6 +36,7 @@ cmd_wait_for_k3s () {
   set -e
 }
 
+# Creates image secrets inside k3s cluster
 cmd_create_image_pull_secret () {
   set -u
   local name=$1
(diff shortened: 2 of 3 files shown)


.buildkite/pipeline.yaml
```diff
@@ -1,9 +1,16 @@
...
       - ./scripts/cibuild test
+  - wait
+  - label: ":shipit: build and release"
+    agents:
+      container_image: ubuntu-2110-ci
+    commands:
+      - ./scripts/cibuild bootstrap
       - ./scripts/cibuild build
       - ./scripts/cibuild export_container "${BUILDKITE_PIPELINE_SLUG}_${BUILDKITE_BUILD_NUMBER}.tar" "docker.greymatter.io/development/gm-operator:latest"
       - ./scripts/cibuild release
```
dontlaugh, line 16: Note that we are still uploading containers to nexus right here, regardless of the outcome of integration tests. 

Not a blocker for this PR, but change this soon.

scripts/cibuild
```diff
@@ -92,40 +101,48 @@ _release_bundle() {
   fi
 }
 
-cmd_generate_integration_pipeline () {
-  # meant to be piped to buildkite-agent
-  echo "
+# Creates the dynamic pipeline that starts up the greymatter mesh environment
+#  in the ephemeral ec2s. The output of the echo commands runs on the ec2, not the original machine running the rest of the pipeline.
+
```
dontlaugh, line 106: remove this newline

scripts/cibuild
```diff
@@ -92,40 +101,48 @@ _release_bundle() {
...
+cmd_generate_integration_tests () {
+  declare steps_yaml
+  cases=('default' 'no_spire' 'tear_down')
 
-steps:
-  - label: \"k3s integration test\"
+  for C in "${cases[@]}"; do
+    launch_cluster $C
+    steps_yaml+=("""
```
dontlaugh, line 113: An alternative to accumulating in an array here would be echoing to a tmp file.

scripts/test
```diff
@@ -16,14 +18,50 @@ test_help () {
...
+_wait_for_resouce() {
+  selector=$1
+  namespace=$2
+  timeout=$3
+  if [ -z $3 ]; then
+    timeout=300
+  fi
+
+  while [ "$($KUBECTL_CMD get -n ${namespace} pods --selector ${selector} -o jsonpath='{.items[*].status.containerStatuses[0].ready}')" != "true" ]; do
+    if [ $timeout -eq 0 ]; then
```
dontlaugh, line 31: would `-lt` be better? if someone changes numbers to a non-multiple of 5 this might break?

scripts/test
```diff
@@ -16,14 +18,50 @@ test_help () {
...
+      echo "Failed while waiting for ${pod}: pod timed out"
+      exit 1
+    fi
+
+    timeout=$(($timeout-5))
+    sleep 5
+  done
+}
+
+# A helper function     Installs the operator with provided tags
```
dontlaugh, line 41: Rewrite this comment to refer to the cue used

scripts/test
```diff
@@ -16,14 +18,50 @@ test_help () {
...
+install_operator () {
+  tags=$1
```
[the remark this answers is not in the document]
dontlaugh: ...where possible. :)

dontlaugh (changes requested):
This is looking fine; The bashisms are complex but ths is a reasonable approach.

Because they're complex, zoom out a little and write a longer comment above each test case that describes the routine.

scripts/test
jack-bischoff: Ah yes great catch

scripts/cibuild
dontlaugh: Not necessarily, but just keep it in mind next time, since it might be easier during the development phase.

danielpcox (approved):
This looks great to me. I love that you're actually checking for connectivity and got rid of the waits, and I love multiple parallel environments for different configurations.

One nit: Could you rewrite it all in Python?

Oh, and `_wait_for_resouce` is misspelled.

dontlaugh (approved):

PR #108, opened 2022-05-02
Pipeline releases only after integration tests succeed.

jack-bischoff:
The CI process shouldn't release binaries or containers if the integration tests (or any tests) fail. This pr fixes that issue by breaking out the release command into its own step, where with the buildkite wait command, will run only if the previous step succeeds.


danielpcox (approved):
I don’t know how other people feel about this, and I’m not asking you to change it now, but I think it would be a good idea to leave the commit that failed the tests, and thereby confirmed the desired behavior, in the commit history as evidence the PR is solid.

PR #110, opened 2022-05-03
Add debug build scripts, config, and instructions

danielpcox:


danielpcox:
By the way, I've only been able to do this on kops so far, because my M1 Mac won't let delve fork off the operator as a subprocess (I think?). I've asked some people in the #delve channel of the Gophers Slack. We'll see if they say anything. https://gophers.slack.com/archives/C07EXM5E0/p1651592755316139

scripts/build
```diff
@@ -15,10 +15,16 @@ cmd_build () {
...
+    #go build -o bin/operator \
+    #    --mod=vendor \
+    #    -tags debug \
+    #    -gcflags 'all=-N -l' github.com/greymatter-io/operator
+    docker build \
+     --platform linux/amd64 \
+     --build-arg username=$GREYMATTER_REGISTRY_USERNAME \
+     --build-arg password=$GREYMATTER_REGISTRY_PASSWORD \
+     -f Dockerfile.debug . -t docker.greymatter.io/internal/gm-operator:debug && \
+    docker push docker.greymatter.io/internal/gm-operator:debug
```
dontlaugh, line 27: Couple of things:

1. Don't push here
2. Let's keep this function for building binaries only, since that's what it does in a few other projects.  Instead, we can extend our existing container building functions. Perhaps a new top-level function:

```sh
cmd_debug_container () {
  if type -p buildah  &>/dev/null; then
    _buildah_build "docker.greymatter.io/internal/gm-operator:latest-debug" Dockerfile.debug
  else
    _docker_build "docker.greymatter.io/internal/gm-operator:latest-debug"  Dockerfile.debug
  fi
}
```

Then add an additional `local` param to `_buildah_build` and `_docker_build` to handle paths to different dockerfiles.

pkg/cuemodule/k8s/outputs/operator.cue
```diff
@@ -205,18 +205,37 @@ operator_sts: [
...
-                path: "/healthz"
-                port: 8081
+            if config.debug {
+              livenessProbe: {
+                httpGet: {
+                  path: "/healthz"
+                  port: 8081
+                }
+                initialDelaySeconds: 120
+                periodSeconds:       20
```
dontlaugh, line 237: Can this liveness probe be hoisted up into the previous if block? E.g. can we get away with just one?

dontlaugh (changes requested):

pkg/cuemodule/k8s/outputs/operator.cue
danielpcox: Yeah, I'll do that

scripts/build
danielpcox: That sounds good to me. I've done as you described. How's it look now?

Dockerfile.debug
```diff
@@ -0,0 +1,56 @@
...
+# args: [
+#   "--listen=:2345",
+#   "--headless=true",
+#   "--log=true",
+#   "--log-output=debugger,debuglineerr,gdbwire,lldbout,rpc",
+#   "--accept-multiclient",
+#   "--api-version=2",
+#   "exec",
+#   "/app/operator",
+# ]
```
dontlaugh, line 56: Add a newline here

dontlaugh (approved):

dontlaugh (approved):

PR #120, opened 2022-05-27
WIP Periodically reconcile Redis listener with environment

danielpcox:
Periodically crawl the watched namespaces (and install namespace) for sidecars, and update the Redis listener so they can beacon their health checks to Redis securely using Spire.


pkg/gmapi/client.go
```diff
@@ -28,8 +28,8 @@ func newClient(operatorCUE *cuemodule.OperatorCUE, mesh *v1alpha1.Mesh, flags ..
 		flags:       flags,
 		ControlCmds: make(chan Cmd),
 		CatalogCmds: make(chan Cmd),
-		ctx:         ctxt,
-		cancel:      cancel,
+		Ctx:         ctxt,
+		Cancel:      cancel,
```
dontlaugh, line 32: Bit of a code smell here. Let's add a method to this struct that derives and returns a child context.

```go
func (c *Client) ContextWithCancel() (context.Context, func()) {
    return context.WithCancel(c.ctx)
}
```

It's good that these functions accept a context, since it provides a handle for cancellation (something you always want with async goroutines) or timeouts (which is also a form of cancellation).

You can use child contexts - derived from a root one like you have already - to delineate scopes of the call stack, and avoid creating channels and signal values of your own.

pkg/gmapi/cli.go
```diff
@@ -142,7 +142,15 @@ func (c *CLI) ConfigureSidecar(operatorCUE *cuemodule.OperatorCUE, name string,
 		logger.Error(err, "Failed to unify or extract CUE", "name", name, "injectedSidecarPort", injectedSidecarPort)
 	}
 
-	ApplyAll(c.client, configObjects, kinds)
+	// Wait for c.Client to exist
+	for {
+		if c.Client != nil {
```
dontlaugh, line 147: It would be better if this check were a method.

```go
if c.Client.Ready() {...}
```
or
```go
ctx = context.WithTimeout(10 * time.Second)
c.Client.Ready(ctx) // internal timeout, error cancelation

select {
  case ctx.Done():
    fmt.Println(ctx.Err())
}
```
I say this _even if_ it delegates to the nil check internally. That's fine for now, but if we encapsulate this in a method we can add more sophisticated readiness checks in one place.

pkg/mesh_install/installer.go
```diff
@@ -246,3 +244,84 @@ func injectGeneratedCertificates(secret *corev1.Secret, cs *cfsslsrv.CFSSLServer
 
 	return secret, nil
 }
+func (i *Installer) reconcileSidecarListForRedisIngress(mesh *v1alpha1.Mesh) {
+	var redisListener json.RawMessage
+	var tempOperatorCUE cuemodule.OperatorCUE
+	var err error
+ReconciliationLoop:
```
alecholmez, line 251: Just a style point I generally see these in all caps:

```suggestion
RECONCILATION_LOOP:
```

dontlaugh: I think GoLand gets mad at caps for this. Which is reason enough for me to not care.

pkg/cuemodule/load.go
```diff
@@ -72,19 +72,24 @@ type Config struct {
 	ClusterIngressName string `json:"cluster_ingress_name"`
 }
 
+type Defaults struct {
```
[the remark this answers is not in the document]
danielpcox: Added a docstring, we'd add members if the Go needed access to anything else in inputs.cue's defaults.

pkg/mesh_install/installer.go
alecholmez: Ah okay I wasn't aware. It was just a stylistic thing so I'm not hard pressed on it

pkg/gmapi/client.go
danielpcox: I'll make an issue and do this in a future PR, since what I think we need instead is a larger refactor to fix the problem that the CLI and greymatter client are two divorced from the "installer". The divisions are obnoxiously arbitrary.

danielpcox: https://app.shortcut.com/grey-matter/story/16384/refactor-divisions-between-operator-internal-components

PR #122, opened 2022-05-31
GitOps watch, fetch, and apply

alecholmez:
[sc-15934]

- Fetches its CUE from a GitOps repository (defaults to [gitops-core](https://github.com/greymatter-io/gitops-core), currently the `update-mesh-upon-changes` branch)
- Polls this repository for changes
- Updates the mesh and injected sidecar configuration upon changes

How to test: Use the updated README in this branch to deploy the operator. The major difference in the README is the necessity of creating a secret for the SSH private key for GitOps.


main.go
```diff
@@ -40,7 +44,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
-	//+kubebuilder:scaffold:imports
+	//_ "net/http/pprof" // DEBUG
```
alecholmez, line 47: Maybe pull this out?

main.go
```diff
@@ -71,11 +83,23 @@ func main() {
...
+	flag.StringVar(&cueRoot, "cueRoot", "core", "Path to the CUE module with Grey Matter config. Defaults to the current working directory.")
 	flag.BoolVar(&zapDevMode, "zapDevMode", false, "Configure zap logger in development mode.")
 	flag.StringVar(&pprofAddr, "pprofAddr", ":1234", "Address for pprof server; has no effect on release builds")
 
+	// Flags that enable bootstrap configuration loading from a git repo.
+	flag.StringVar(&bootstrapRepo, "repo", "", "Bootstrap repository for operator configuration.")
+	flag.StringVar(&bootstrapSSHKeyPath, "sshPrivateKeyPath", "", "SSH key which has privileges to fetch the operators core configuration from Git.")
+	flag.StringVar(&bootstrapSSHKeyPassword, "sshPrivateKeyPassword", "", "Password for the SSH key")
+	flag.StringVar(&bootstrapBranch, "branch", "main", "target branch to fetch and watch for changes in the core configuration repo.")
+	flag.IntVar(&interval, "interval", 30, "Interval to watch bootstrap core config repo.")
```
alecholmez, line 101: Can we namespace this variable too so we know it’s just for sync

pkg/mesh_install/installer.go
```diff
@@ -148,8 +153,34 @@ func (i *Installer) Start(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+		// copy in old mesh dynamic values
+		freshLoadMesh.TypeMeta = i.Mesh.TypeMeta
+		i.Mesh.ObjectMeta.DeepCopyInto(&freshLoadMesh.ObjectMeta)
+
+		i.ApplyMesh(i.Mesh, freshLoadMesh)
+
+		//i.OperatorCUE = freshLoadOperatorCUE
```
alecholmez, line 170: Can we delete this commented code?

pkg/mesh_install/installer_test.go
```diff
@@ -11,6 +11,26 @@ import (
 )
 
 func TestLoading(t *testing.T) {
+	// memory leak test
```
alecholmez, line 14: Delete this?

pkg/mesh_install/installer_test.go
```diff
@@ -211,11 +231,11 @@ func TestLoading(t *testing.T) {
 	//t.Logf("New value: %#v", newValue)
 
 	// UNMARSHALL TEST
-	data := []byte("{\"mesh_id\":\"Experiment\"}")
-	var extracted struct {
-		MeshID string `json:"mesh_id"`
-	}
-	_ = json.Unmarshal(data, &extracted)
-	t.Logf("meshid:%v", extracted.MeshID)
+	//data := []byte("{\"mesh_id\":\"Experiment\"}")
```
alecholmez, line 234: Delete this?

pkg/sync/sync.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	return s
+}
+
+// WithSSHInfo will set a users ssh information on sync config.
+// Passwords are not required.
+func WithSSHInfo(privateKeyPath, password string) func(*Sync) {
+	//var privateKey string
+	//if privateKeyPath != "" {
```
alecholmez, line 55: Do we need?

alecholmez (commented):
Just some small things here and there and then should be good to merge. 

I’m still hesitant to merge this before we update gitops core

main.go
danielpcox: I left it in on purpose, so someone could just uncomment if they felt like getting pprof, but I'm not super attached to it.

pkg/sync/sync.go
danielpcox: I'll remove this

pkg/mesh_install/installer_test.go
danielpcox: Uh... I'd rather not? I use this all the time to just try stuff out... If I were in Python, this would be a Jupyter notebook that I use for experimentation. I know it's not a thing Go programmers usually do, but... maybe they should? leave themselves some tooling lying around they find helpful when they want to debug something...?

danielpcox: Help me think about this, from a Go perspective. Would you add something like this to prevent it from being built, and to tell users that this isn't normally used, and yet it has some uses?
```
//go:build debug
// +build debug
```

pkg/mesh_install/installer_test.go
danielpcox: I'd rather not. See below.

pkg/mesh_install/installer.go
danielpcox: Done ✅

main.go
danielpcox: Good plan. I've renamed everything with "bootstrap" in it to have "sync" instead, and renamed "interval" to "syncInterval".

pkg/mesh_install/installer_test.go
alecholmez: I think that’s a reasonable, assumption to only build that when it’s in a debug build.

PR #125, opened 2022-06-06
WIP: Upstream updates and gen cuemodule

rileyp2000:


tilleryd:
@danielpcox @dontlaugh We determined that some of these changes are from 3 years ago in https://github.com/kubernetes/api/blame/master/apps/v1/types.go. The trigger for these changes were caused by a more recent version of cue-cli. So we should either lock down the version of cue-cli or k8s api. I don't know which is best so please chime in.

dontlaugh:
I would say that pinning `cue` is more important, because it's going to have the more unstable API for the time being.

This means tracking releases and ensuring that CI containers have the right version.

danielpcox:
I'd be very afraid to merge this... I think we should close the PR for now, wait for the dust to settle on recent and incoming big changes to `main`, and then re-run and test thoroughly. I don't think the conflicts are likely to be manually reconcilable on this branch as it is right now.

tilleryd:
This is languishing and seems to not be an issue at the moment. Closing.

PR #134, opened 2022-07-14
Upgrade packaged CLI to 4.2

danielpcox:
[sc-19021]

This turned out to be quite straightforward, despite some unrelated weirdness holding things up. Two changes needed to be made in the operator to support the 4.0.1 -> 4.2.0 upgrade, other than the upgrade itself:

- "host" -> "url" in the embedded config.toml
- use error codes instead of a response containing "200"

If the build succeeds, this PR is good to go, since the integration tests exercise the mesh config deployment code including using the new CLI version.

### DIFF
--- a/pkg/gmapi/commands.go
+++ b/pkg/gmapi/commands.go
@@ -3,8 +3,6 @@ package gmapi
 import (
 	"encoding/json"
 	"fmt"
-	"strings"
-
 	"github.com/tidwall/gjson"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/gmapi/commands.go
```diff
@@ -21,13 +19,6 @@ func MkApply(kind string, data json.RawMessage) Cmd {
 				logger.Info("apply", "type", kind, "key", key)
 			}
 		},
-		modify: func(out []byte) ([]byte, error) {
```
alecholmez, line 24: No longer need this?

danielpcox: I'm going to leave it as an option in case we need it in the future, but it's not strictly part of the current use case, no. It's for optionally changing the definition of success and failure wrt CLI commands.

alecholmez: Ah great! Sounds good

alecholmez (approved):

PR #136, opened 2022-07-19
update gitops-core submodule

alecholmez:
Signed-off-by: Alec Holmes <alecholmez@me.com>


danielpcox:
Closing this because I'm about to do basically the same thing, with an even more recent version.

PR #142, opened 2022-08-10
Support tags for clone and watch/update

danielpcox:
[sc-21671]

Adds a `-tag` argument to the operator binary, mutually exclusive with `-branch`, which watches a tag of the gitops repository for updates to apply to the environment. Force-pushing an update to the given tag name results in a fetch and checkout of the new tag followed by an environment update.


main.go
```diff
@@ -97,7 +98,8 @@ func run() error {
 	flag.StringVar(&syncRepo, "repo", "", "Bootstrap repository for operator configuration.")
 	flag.StringVar(&syncSSHKeyPath, "sshPrivateKeyPath", "", "SSH key which has privileges to fetch the operators core configuration from Git.")
 	flag.StringVar(&syncSSHKeyPassword, "sshPrivateKeyPassword", "", "Password for the SSH key")
-	flag.StringVar(&syncBranch, "branch", "main", "target branch to fetch and watch for changes in the core configuration repo.")
+	flag.StringVar(&syncTag, "tag", "", "target tag to fetch and watch for changes in the core configuration repo.")
+	flag.StringVar(&syncBranch, "branch", "", "target branch to fetch and watch for changes in the core configuration repo. defaults to 'main' if no branch or tag specified")
```
alecholmez, line 102: We probably want to have a default branch value here right? Otherwise if they don't give a tag it won't know what to pull or will it just pull HEAD?

main.go
```diff
@@ -113,7 +115,7 @@ func run() error {
 	// build sync options based on user configuration.
 	syncOpts := []func(*sync.Sync){}
 	syncOpts = append(syncOpts, sync.WithSSHInfo(syncSSHKeyPath, syncSSHKeyPassword))
-	syncOpts = append(syncOpts, sync.WithRepoInfo(syncRepo, syncBranch))
+	syncOpts = append(syncOpts, sync.WithRepoInfo(syncRepo, syncBranch, syncTag))
```
alecholmez, line 118: What happens when these are both empty strings?

pkg/sync/sync.go
```diff
@@ -59,10 +60,21 @@ func WithSSHInfo(privateKeyPath, password string) func(*Sync) {
 
 // WithRepoInfo will set target repository information
 // on a sync configuration object.
-func WithRepoInfo(remote, branch string) func(*Sync) {
+func WithRepoInfo(remote, branch string, tag string) func(*Sync) {
+	// If neither a branch nor a tag is specified, default to the main branch
+	if branch == "" && tag == "" {
+		branch = "main"
+	}
```
alecholmez, line 67: Ah I see now. Should we move move this into main? I feel as if its a little buried in here so it's not immediately obvious when looking at the top level initialization steps of the operator.

pkg/sync/sync.go
```diff
@@ -59,10 +60,21 @@ func WithSSHInfo(privateKeyPath, password string) func(*Sync) {
...
+
+	// You cannot specify both a branch and a tag
+	if branch != "" && tag != "" {
+		panic("You must specify a branch OR a tag for GitOps, not both")
```
alecholmez, line 71: Do we want a panic() or do we just want to error

alecholmez (changes requested):

pkg/sync/sync.go
danielpcox: I think we need it to die hard right here, 'cause we can't continue and it's going to die anyway. Might as well do it with a helpful error message.

pkg/sync/sync.go
danielpcox: Yeah, I'll do that.

pkg/sync/sync.go
alecholmez: I think my thought was we don't need a super complicated stack trace failure from a `panic()`. We could even do something like `log.Fatal("You must specify a branch OR a tag for GitOps, not both")` so it dies hard but it's easily parseable instead of getting a big scary panic.

danielpcox: Works for me

alecholmez (approved):
Nice let's get this release out the door

PR #143, opened 2022-08-11
WIP: Use buildkite to launch ephemeral k3s boxes

dontlaugh:


dontlaugh:
Our SaaS provider fixed their bug, so we don't need to merge this immediatlely

PR #147, opened 2022-08-12
Add optional Bugsnag error reporting using go build flags

jack-bischoff:
Bugsnag at best, is irrelevant to end users, and at worst,
may introduce additional attack vectors, get flagged in
software scans, have network traffic blocked entirely, or
just generally cause pause for customers. To combat these
issues, we need to create separate dev/prod builds.
The dev build will contain the library so we can collect
error metrics during operation in test environments.
The production release will not only disable bugsnag
but will not even contain bugsnag code at all, so that
potential supply chain scans won't flag it.

The solution lies in go build flags and an error
reporter "interface" abstraction. Operator code interacts
with bugsnag library functions through the reporter
package functions. These functions essentially
shadow the bugnsag functions of the same name.
However, this allows us to conditionally compile
empty function stubs instead of the normal
definitions during non-dev builds. Go includes the
empty function stubs if it does not detect a build tag
and includes the working functions if it detects the
"dev" build tag. We pass in this tag through Docker
args.

Lastly, to support this new build pathway, the ci/cd
process expanded and changed to build both images,
push both images to quay, and run the dev image
during integration testing.

Other things to note: I included the CA certs for bugsnag's cert into the operator image as the base ubuntu image doesnt ship with any. I also corrected the pipeline agents to the 22.04. I modified the README to include instructions on how to use the dev version

To test:
First, install the operator without bugsnag (the non-dev latest version). Check the logs to ensure that `Initializing bugsnag` doesn't appear.

Then, follow the instructions in the readme to install the dev version of the operator.  Ensure that the logs contain `Initializing bugsnag`


dontlaugh:
@alecholmez Ask your boy how to write commit messages.

main.go
```diff
@@ -76,19 +76,14 @@ var (
...
 func main() {
+	reporter.Init()
 	if err := run(); err != nil {
 		logger.Error(err, "Failed to run operator")
 		os.Exit(1)
 	}
 }
 
 func run() error {
-	defer func() {
```
dontlaugh, line 86: We do not need to catch panics in this codebase, so I support removing this. Better to crash, since operators will be running on k8s anyways.

pkg/reporter/empty_reporter.go
```diff
@@ -0,0 +1,6 @@
+//go:build !dev
+
```
dontlaugh, line 2: If the `reporter` package is a bugsnag-specific interface, let's give it a bugsnag-specific name.

dontlaugh (changes requested):
This distinction is indeed required, but a "debug build" is a more accurate name, in my opinion.

Can we fold this into our existing debug build?

pkg/reporter/reporter.go
```diff
@@ -0,0 +1,27 @@
...
+
+package reporter
+
+import (
+	"fmt"
+	"github.com/bugsnag/bugsnag-go/v2"
+	"os"
+)
+
+func Init() {
```
alecholmez, line 11: A Doc string on the `Init()` function about the bugsnag functionality would be nice.

pkg/reporter/reporter.go
```diff
@@ -0,0 +1,27 @@
...
+		fmt.Println("Initializing Bugsnag")
+		bugsnag.Configure(bugsnag.Configuration{
+			APIKey:          t,
+			ReleaseStage:    "development",
+			ProjectPackages: []string{"main", "github.com/greymatter-io/operator/pkg"},
+		})
+	}
+}
+
+func Notify(err error, rawData ...interface{}) error {
```
alecholmez, line 25: Docstring here too since it's exported

alecholmez (changes requested):

dontlaugh (commented):
Squash again when this is ready

pkg/bugsnag/reporter.go
```diff
@@ -0,0 +1,36 @@
...
+			APIKey:          t,
+			ReleaseStage:    "development",
+			ProjectPackages: []string{"main", "github.com/greymatter-io/operator/pkg"},
+		})
+	}
+}
+
+//Notify shadows bugsnag's Notify function which sends an error to the api.
+//https://godoc.org/github.com/bugsnag/bugsnag-go/v2/#Notify
+func Notify(err error, rawData ...interface{}) error {
```
alecholmez, line 34: Just for safekeeping we should probably instrument this through the operator code. No action here but worth a ticket in the future.

alecholmez (changes requested):

dontlaugh:
What else should we be looking for here? Can you email the team with some links to dashboards, reports, etc.? Have you tested with a pre-release build to see if you've integrated bugsnag successfully?

jack-bischoff:
> What else should we be looking for here? Can you email the team with some links to dashboards, reports, etc.? Have you tested with a pre-release build to see if you've integrated bugsnag successfully?

@dontlaugh 

Yep, I'll email the team. The free bugsnag plan only allows for 1 account. I put the account information in vault, anyone can log in until we decided if/when to upgrade. I have tested it and bugsnag does report home. It only keeps logs for 7 days on the free plan so I manually triggered some errors today so they appear. Additionally, you can log into jack.k8s.local to verify that the debug-latest build works.

alecholmez (dismissed):

.buildkite/pipeline.yaml
```diff
@@ -25,6 +29,7 @@ steps:
       container_image: ubuntu-2204-ci
   - wait
   - label: "release"
+    if: "build.branch == 'main' || build.tag != null"
```
dontlaugh, line 32: nice

scripts/test
```diff
@@ -45,9 +45,12 @@ _wait_for_resource() {
...
 install_operator () {
-  local tags=$1
-  local image_tag="${BUILDKITE_PIPELINE_SLUG}_${BUILDKITE_BUILD_NUMBER}"
-  ( cd pkg/cuemodule/core && cue eval -c ./k8s/outputs -t operator_image="docker.greymatter.io/development/gm-operator:$image_tag" $tags  --out text -e operator_manifests_yaml | $KUBECTL_CMD apply -f - )
+  local tags="$1"
+  local image_tag="${BUILDKITE_PIPELINE_SLUG}_${BUILDKITE_BUILD_NUMBER}_debug"
+  cd pkg/cuemodule/core || exit
+  cue eval -c ./k8s/outputs -t operator_image="quay.io/greymatterio/operator:$image_tag" -t test=true $tags  --out text -e operator_manifests_yaml | $KUBECTL_CMD apply -f -
+  cd - || exit
+
```
dontlaugh, line 53: We usually use subshells for all `cd`; We can try your way, but clean up the comments around this function, and break the `cue eval` onto a couple of lines.

dontlaugh (changes requested):
Cleanup request in `install_operator` function.

alecholmez (approved):
Once the builds are green this is good from my end

Dockerfile.debug
```diff
@@ -18,36 +18,37 @@ ARG username
...
+ENV SSH_KNOWN_HOSTS=/app/known_hosts
 
 WORKDIR /app
+COPY --from=builder /root/.ssh/known_hosts /app/known_hosts
 COPY --from=builder /workspace/operator /app/operator
 COPY --from=builder /go/bin/dlv /app/dlv
 COPY --from=builder /workspace/greymatter /bin/greymatter
 COPY --from=builder /workspace/pkg/cuemodule/core /app/core
-RUN mkdir .config && chmod 777 .config
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
```
dontlaugh, line 52: Nice touch. But I'm a bit worried about the crazy looking URLs in the previous build going stale. Not sure if there's anything better to do about it right now.

We might consider maintaining a "base" image for these debug builds, if there's going to be common assets like this. Are those certs specific to this project, or can they be shared across our projects? 

cc @johnworthing @clbx @mikeschear @joshuaswickirl @moosecanswim

dontlaugh (changes requested):
Looks good, but we need a CHANGELOG update for this.

dontlaugh (approved):
Approved. There are integration test failures, though, and that's a bummer. But it seems to be flaky tests. I'll look into it soon.

PR #152, opened 2022-08-18
Avoid top-level agent targeting

dontlaugh:
If we use a top-level agent target, it affects subsequent steps in unexpected ways.
In this case, specifically, the *generated* steps were affected, and had the
`container_image` agent tag set. The virtual machines that get launched do
not have this parameter set, so the generated steps were waiting forever for 
a VM-based agent that would never come online.


danielpcox (approved):
Let's try it

PR #159, opened 2022-08-26
Solve bootstrap command ordering bug

jack-bischoff:
The bootstrap script will attempt to pull a cuemodule down before
any submodules are fetched. If a user clones the repo without
initializing submodules, this will result in an error.
The solution is a matter of re-ordering checks. By putting git
as the first dependency check, submodules will get created
before entering those directories.


dontlaugh (approved):
This is good; although given that this fetch can actually manipulate the checked-in files on disk in unexpected ways, it might not belong in bootstrap at all.

PR #160, opened 2022-09-01
Preserve cluster and workload labels if already present

danielpcox:
[sc-24122]

Only set either the `greymatter.io/cluster` or `greymatter.io/workload` label if one doesn't already exist.

### How to test

1. Build this branch
2. Launch the operator
3. Wait for a mesh
4. Deploy a workload that receives an injected sidecar, but that already has a cluster label and workload label set to something arbitrary, such as:
```
spec:
  template:
    metadata:
      labels:
        "greymatter.io/cluster": "something-arbitrary
        "greymatter.io/workload": "equally-arbitrary"
    
```
5. Confirm that the workload, even after being modified by the operator to inject a sidecar, does not modify these labels.
6. Delete your workload.
7. Deploy another workload without these labels, and confirm that the labels are automatically added. 

Note: The sidecar with your manually-added labels is not expected to _work_, even if you add the configure-sidecar annotation, because your manually-created label won't match what it needs to for service discovery to work. The real-world use-case of this work is to be able to take manual control of both sides of this equation and do the Grey Matter configuration manually.


joel-epstein (approved):
I have tested this PR using the Test Plan provided.  Cluster and Workload labels can be manually applied by a User and are also applied by the operator upon discovery of a workload if not provided by the User.

PR #161, opened 2022-09-01
Make release_version optional in the Mesh definition

danielpcox:
[sc-14772]

Companion PR in gitops-core: https://github.com/greymatter-io/gitops-core/pull/68
(which needs to be merged along with this one)

This PR makes the `release_version` optional (and default-omitted) in the Mesh definition.

### How to test

1. Point the submodule at the companion PR branch in GitOps-core
2. Build this branch.
3. Launch an operator without any `mesh.release_version` set in inputs.cue. If it works at all, this PR has succeeded in its purpose, because previously it would fail to load the CUE without `mesh.release_version` present in inputs.cue.


joel-epstein (approved):
Operator successfully installs and Mesh deployed without mesh.spec.release_version KV in inputs.cue.  Looks good.

PR #164, opened 2022-09-06
fix ssh key name

tilleryd:


pkg/gitops/sync_test.go
```diff
@@ -89,7 +89,7 @@ func TestNewSyncOpts(t *testing.T) {
 func TestSyncLifecycle(t *testing.T) {
 	// get ssh key path - right now this looks for an
 	// ecdsa key due to github deprecating rsa keys support.
-	path := filepath.Join(os.Getenv("HOME"), ".ssh", "id_ed25519")
+	path := filepath.Join(os.Getenv("HOME"), ".ssh", "ssh_private_key")
```
dontlaugh, line 92: This [fails the tests, but for a bad reason](https://buildkite.com/greymatter/operator/builds/1089#01831409-4271-4f21-84b9-4ed44b8526c3). It relies on the naming of a file under $HOME, we can't assume that.

dontlaugh (changes requested):
Let's back out the changes in **sync_test.go**; Regarding depending on the local file, I'll make a note to follow up later this week in my operator integration test review.

dontlaugh:
The markdown changes are fine, and I _think_ the cibuild script changes are fine, because it relies on the same key, but just changes the name of the secret's field.

pkg/gitops/sync_test.go
tilleryd: Done, thanks.

dontlaugh (approved):

PR #165, opened 2022-09-06
Support external spire

danielpcox:
[sc-13415]

Companion PR in gitops-core: https://github.com/greymatter-io/gitops-core/pull/73


pkg/cuemodule/load.go
```diff
@@ -75,6 +76,11 @@ type Config struct {
 	OperatorNamespace  string `json:"operator_namespace"`
 }
 
+type Spire struct {
```
dontlaugh, line 79: This holds the config of a K8s secret's 1) name and 2) namespace, right? Should we call this (something like) `SpireSecret`?

In any case, add a doc string to all exported members in Go files.

danielpcox: No, it's not actually a secret, and that's not all it holds. It's a miscellaneous bag of stuff necessary to point at an existing spire installation, and the actual block is named `defaults.spire` so I'd prefer to call it "spire" to match. It doesn't need to be exported though, so I lowercased it, and I added a doc string to the parent `Defaults` struct.

dontlaugh: unexporting it is even better

dontlaugh (dismissed):

dontlaugh (approved):

dontlaugh:
Good to merge if the tests are green.

PR #166, opened 2022-09-06
Do not echo bash commands by default

dontlaugh:
Selectively set -x if you need to, or only use while debugging.


dontlaugh:
_Always_ setting `-x` is a bit noisy, imo

jack-bischoff (approved):

PR #167, opened 2022-09-08
CUE loading unit tests

danielpcox:
[sc-22033]


pkg/cuemodule/load_test.go
```diff
@@ -1,40 +1,157 @@
...
 	}
+	ctx = context.Background()
+}
 
-	t.Run("TestExtractConfig", func(t *testing.T) {
-		config, _ := operatorCUE.ExtractConfig()
+func TestLoad(t *testing.T) {
+	config, defaults := operatorCUE.ExtractConfig() // FUT
 
+	t.Run("TestExtractedConfigAndDefaults", func(t *testing.T) {
```
alecholmez, line 45: It looks like you're executing quite a few tests inside of TestLoad. Could these be converted to table driven tests?

https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

danielpcox: How about we do that later so we can get these tests in? I don't offhand remember right now if we could have used table-driven tests or not for this, but I don't think we're actually _changing_ any inputs in this set of tests, just confirming several different things from the same input.

pkg/cuemodule/load.go
```diff
@@ -271,13 +272,15 @@ type justKeys struct {
 	CatalogZoneKey   string `json:"zone_key"`   // CatalogService
 }
 
+var unrecognizedKindLabel = "unrecognized"
```
alecholmez, line 275: Can this be a `const`?

pkg/cuemodule/load_test.go
alecholmez: Okay no worries. I just read through it and I think this is okay.

pkg/cuemodule/load_test.go
```diff
@@ -1,40 +1,160 @@
...
+		defaults.SidecarList = []string{"foo", "bar"}
+		tempOperatorCUE, err := operatorCUE.TempGMValueUnifiedWithDefaults(defaults) // FUT 1
+		if err != nil {
+			t.Fatal(err)
+		}
+		redisListener, err := tempOperatorCUE.ExtractRedisListener() // FUT 2
+		if err != nil {
+			t.Fatal(err)
+		}
+		//t.Fatal(string(redisListener))
```
alecholmez, line 68: Delete this?

pkg/cuemodule/load_test.go
```diff
@@ -1,40 +1,160 @@
...
+		assert.Equal(t, expectedSpireBlock, defaults.Spire)
+	})
+
+	t.Run("Test defaults injection for sidecar list", func(t *testing.T) {
+		defaults.SidecarList = []string{"foo", "bar"}
+		tempOperatorCUE, err := operatorCUE.TempGMValueUnifiedWithDefaults(defaults) // FUT 1
+		if err != nil {
+			t.Fatal(err)
+		}
+		redisListener, err := tempOperatorCUE.ExtractRedisListener() // FUT 2
```
alecholmez, line 64: Not sure what this means:
```
// FUT 2
```

pkg/cuemodule/load_test.go
```diff
@@ -1,40 +1,160 @@
...
+			Namespace:    "spire",
+			TrustDomain:  "greymatter.io",
+		}
+		assert.Equal(t, expectedSpireBlock, defaults.Spire)
+	})
+
+	t.Run("Test defaults injection for sidecar list", func(t *testing.T) {
+		defaults.SidecarList = []string{"foo", "bar"}
+		tempOperatorCUE, err := operatorCUE.TempGMValueUnifiedWithDefaults(defaults) // FUT 1
+		if err != nil {
```
alecholmez, line 61: Wherever you're checking for `if err != nil` lets use assert instead:
```go
assert.NoError(t, err)
```

pkg/cuemodule/load_test.go
```diff
@@ -1,40 +1,160 @@
...
+		}
+		redisListener, err := tempOperatorCUE.ExtractRedisListener() // FUT 2
+		if err != nil {
+			t.Fatal(err)
+		}
+		//t.Fatal(string(redisListener))
+		res := gjson.GetBytes(redisListener, "secret.subject_names")
+		subjects := res.Array()
+		for i, clusterName := range defaults.SidecarList {
+			expectedSubject := helperSpireSubject(
```
alecholmez, line 72: It's generally idiomatic go to do this format instead:
```go
			assert.Equal(
				t,
				helperSpireSubject(
				    defaults.Spire.TrustDomain,
				    config.OperatorNamespace,
				    mesh.Name,
				    clusterName,
                ),
				subjects[i].String(),
            )
```

alecholmez: If you'd like to keep the separate variable that also works too but the real nitpick I have is this `)` on the final line of the function call needs its own separate line

pkg/cuemodule/load_test.go
```diff
@@ -1,40 +1,160 @@
...
+			if env.Name == "XDS_CLUSTER" {
+				xdsClusterExists = true
+				assert.Equal(t, env.Value, clusterLabel, "XDS_CLUSTER label does not match clusterLabel")
+			}
 		}
-		tempOperatorCUE, _ := operatorCUE.TempGMValueUnifiedWithDefaults(defaults)
-		redisListener, _ := tempOperatorCUE.ExtractRedisListener()
-		logger.Info("Loaded Redis listener", "listener", redisListener)
-		//logger.Info("LoadAll sidecarList", "SidecarList", defaults.SidecarList)
+		if !xdsClusterExists {
```
alecholmez, line 115: This can also be an assert

pkg/cuemodule/load_test.go
```diff
@@ -1,40 +1,160 @@
...
+				//fmt.Println(res.String())
+				assert.Contains(t, resCluster.String(), sidecarName)
+			}
+			resPort := gjson.GetBytes(o, "instances.0.port")
+			if resPort.Int() != 0 {
+				sidecarPortExists = true
+				assert.Equal(t, int64(sidecarPort), resPort.Int())
+			}
+		}
+		if !sidecarPortExists {
```
alecholmez, line 146: Assert?

alecholmez (changes requested):
Just some minor things but overall tests look great

pkg/cuemodule/load_test.go
danielpcox: Function Under Test

alecholmez (approved):
Looks good!

danielpcox:
Oh hey, nice job @joel-epstein! Please do the honors.

PR #170, opened 2022-09-09
WIP Unit tests for mesh_install package

danielpcox:
[sc-22035]


pkg/gitops/state_backup.go
```diff
@@ -297,3 +297,16 @@ func (ss *SyncState) persistSidecarHashesToRedis(hashes map[string]SidecarObject
 		logger.Error(err, "Failed to save sidecar environment state hashes to Redis", "hashes", hashes)
 	}
 }
+
+func NewMockSyncState() *SyncState {
```
alecholmez, line 301: I would actually move this into the `Testing` go file. It's generally good practice to keep mocks and testing code out of the original functionality files.

pkg/gmapi/cli.go
```diff
@@ -170,3 +187,62 @@ func (c *CLI) UnconfigureSidecar(operatorCUE *cuemodule.OperatorCUE, name string
 
 	UnApplyAll(c.Client, configObjects, kinds)
 }
+
+// MockCLI implements CLI for testing purposes
```
alecholmez, line 191: Same thing with this. Can we move this into `cli_test.go` file in the same package?

alecholmez (changes requested):
This actually looks great nice job!

Just 2 minor requests from me

pkg/gitops/state_backup.go
danielpcox: I'm skeptical of this "it's generally good practice", because it seems to me that it's much better to keep this code close to the SyncState struct definition. That way all of the code that needs to know the internals of SyncState can stay in SyncState. I also may (probably will, actually) need to use this function from other test files. You OK if I leave it here?

pkg/gmapi/cli.go
danielpcox: I _definitely_ want to leave this one here. If it's moved elsewhere, it's far away from the other implementations of this interface, which is part of the reason I find the Control code so hard to read.

danielpcox:
I wouldn't mind moving these into separate files, but I wouldn't want to move them into the tests that use them, 'cause IMO they belong with the interface they're implementing.

danielpcox:
I've moved them into `_mock.go` files to satisfy the Go gods.

alecholmez (approved):

PR #173, opened 2022-09-13
Remove "bundle" generation from the release process

danielpcox:
The OpenShift OLM GUI installation of operators requires a "bundle" container built, but we haven't done that since the big operator refactor and this is failing builds. Removing it from the release processs for now.


dontlaugh:
This is worth a changelog update

dontlaugh (approved):

PR #175, opened 2022-09-21
Maintainable integration tests

danielpcox:
[sc-22225]

So far I've only rewritten `scripts/test` in Python, in a way that essentially tracks the original bash script, plus a few functional bugfixes and improvements. In #176 I'll be taking this further.

This is probably not its final organization, but IMO we should merge it like this anyway and I'll continue to improve it.


scripts/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+    return result
+
+
+def install_operator(args, env, tags):
+    """Installs the operator assuming cue and kubectl (kubectl is overrideable in args for meta-tests)"""
+    image_tag = f'{env.PIPELINE_SLUG}_{env.BUILD_NUMBER}_debug'
+    heading(f'Installing the operator (:{image_tag})')
+    print('Generating manifests using CUE tags:', tags)
+
+    run(f'''(
```
dontlaugh, line 71: `run` is the name of a function in `subprocess` module. Can we call our own version of this `run_shell` or `runsh` or something?

scripts/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+    cue eval -c ./k8s/outputs --out text -e operator_manifests_yaml \
+        -t operator_image="quay.io/greymatterio/operator:{image_tag}" \
+        {tags} | {'cat' if args.dryrun else 'kubectl apply -f -'}
+    )''')
+
+
+def wait_for_pod(args, selector, namespace='greymatter', retries=60, interval=5):
+    """Waits on a Pod in Kubernetes (blocks)"""
+    if args.dryrun:
+        print(f'dryrun is set, but would otherwise wait for pod "{selector}" in namespace {namespace}')
```
dontlaugh, line 82: Maybe we don't need this log message. Do you expect this to get printed a lot in our CI output?

scripts/helpers.py
```diff
@@ -0,0 +1,173 @@
...
+        return wrap
+
+    return dec
+
+
+@retry(5)
+def upload_operator_logs(args, operator_namespace='gm-operator'):
+    """Robustly upload the operator logs to buildkite artifact storage"""
+    if args.dryrun:
+        print(f'dryrun is set, but would otherwise upload operator logs')
```
dontlaugh, line 124: ditto here

dontlaugh (changes requested):
Let's rename the `run` function, that's my only blocker to merging this.

scripts/helpers.py
danielpcox: That was intentional, because my `run` takes exactly the same keyword args as `subprocess.run`, just with different defaults. Do you still think we should change it?

scripts/helpers.py
danielpcox: Not at all. It's for meta-testing. It's just if you want to run the test script to see what it would do given certain inputs, without it actually performing any actions. I also intend to refactor this into `heading` or a decorator so it's not so repetitive.

dontlaugh (approved):

dontlaugh (approved):

danielpcox:
After a conversation, we'll leave it run for now, and change it to shell in my next PR.

joel-epstein (commented):
I took a look.

PR #176, opened 2022-09-22
GitOps-driven tests, and adding Observables tests

danielpcox:
[sc-22225]

#175 rewrote `scripts/test` in Python, and this PR is about adding functionality that wasn't present in the bash script version - Running tests with GitOps, and tests for the observability pipeline.

Also renaming my `run` to `shell` to avoid accidental future collision with `subprocess.run`.


scripts/test
```diff
@@ -7,62 +7,56 @@ from pathlib import Path
...
 
-# Handler for the lint subcommand
 def lint_handler(args):
-    h.run('staticcheck ./...')
+    """Handler for the lint subcommand"""
+    h.shell('staticcheck ./...')
 
 
-# Handler for the fast subcommand
 def fast_handler(args):
```
dontlaugh, line 15: My only concern with this script is the identification of `go test` with "fast" will increase the risk that someone adding a "slow" variant will skip them.

dontlaugh (approved):

scripts/test
danielpcox: Haha yeah. I didn't pick the name, just kept it from the old integration tests. I thought it was pretty odd too. I would have called them `unit`.

PR #177, opened 2022-09-23
use greymatter-core repo

tilleryd:
@danielpcox In order to get this to work, I had to run `cd pkg/cuemodule/core; git remote set-url origin git@github.com:greymatter-io/greymatter-core.git; git submodule update --recursive --remote`.


pkg/gitops/sync_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	gitRemote = "git@github.com:greymatter-io/gitops-core.git"
+	gitRemote = "git@github.com:greymatter-io/greymatter-core.git"
```
danielpcox, line 14: We might want to leave this one as it was, because the test of sync should probably break if the bleeding-edge gitops-core repo does something to break it.

PR #178, opened 2022-09-26
Multi-operator deployment tests

danielpcox:
[sc-27235]

In a third EC2 running k3s, deploys two operators into separate namespaces, and confirms that the resulting meshes are correctly deployed and reachable.


scripts/helpers.py
```diff
@@ -198,48 +245,55 @@ def wait_for_pod(args, selector, namespace='greymatter', retries=60, interval=5)
...
+@timing
 @retry(24, interval=10)
 def curl(url):
     """Execute quiet curl with retry on failure"""
     return shell(f'curl -4 -sSLf -o /dev/null {url}').stdout
 
 
+@threaded
+@retry(10)
+def port_forward(namespace, id, ingress_port, egress_port):
```
danielpcox, line 291: I reeeeeeeally wanted to make `port_forward` into a context manager so I could do something cool like this:

```
with h.port_forward(namespace='second', id='deploy/edge', ingress_port=65432, egress_port=10808):
    h.confirm_mesh_reachable(args, env, port=65432)

```

but alas I couldn't figure out how to use subprocess to both to retry on errors _and_ give me the process to be killed once the context manager's yield completed in a reasonable amount of time, so we have to settle for this only _kinda_ cool `@threaded` decorator.

scripts/fixtures/second_operator.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/inputs.cue b/inputs.cue
```
dontlaugh, line 1: Cane we add a comment about how this works? I believe some explanatory text can be added above this unified diff, and it should still apply cleanly.

scripts/helpers.py
```diff
@@ -93,32 +94,83 @@ def shell(command, **kwargs):
...
         raise e
 
+
+def threaded(f):
+    """Decorator to run a function call in a separate thread"""
+
+    @wraps(f)
+    def wrap(*args, **kwargs):
+        t = Thread(target=f, args=args, kwargs=kwargs, daemon=True)
+        t.start()
```
dontlaugh, line 104: Our usage of this for port forwarding needs to be non-blocking, but I'm not thrilled that there's no way to explicitly cancel this. In practice it doesn't seem to matter _now_, and we intend to rewrite this in Go rather soon, but consider my misgivings registered! :scroll:

scripts/helpers.py
dontlaugh: Yeah :( I think we're missing an object here to manage this stuff. Not a blocker for me tho

dontlaugh (changes requested):
Document the patch file where I've left my comment, and we'll be good. My other comments are advisory only.

scripts/helpers.py
danielpcox: Yeah, that's exactly why I wanted to do it with a context manager so that it would only exist for the short time that we needed it. I'll try to figure it out again later if I have some time.

danielpcox:
I just put comments on all of the patch files. We'll see if that actually works.

scripts/helpers.py
danielpcox: Oh, I thought we were going to call the Python integration tests from Go.

danielpcox:
Looks like it worked just fine. 🎉

dontlaugh (approved):

PR #181, opened 2022-09-28
Removing --remote from git submodule update commands

danielpcox:
This command causes all of the submodules of the current repository (recursively) to be updated to the pinned version that the current commit has baked in:

```
git submodule update --init --recursive
```

This command, on the other hand, causes all of the submodules of the current repository (recursively) to be updated to the very latest tip of the branch specified in `.gitmodules`, which is really dangerous - you may think you're testing with one dependency, when you're really testing with the very latest. It's like using `*` in your dependencies list instead of pinning a version:

```
git submodule update --init --recursive --remote
```

This PR removes all of the `--remote`s from `git submodule update` commands throughout the operator repository.


danielpcox:
Tagging lots of people just for the FYI

dontlaugh:
Good call :+1:

dontlaugh (approved):

joel-epstein (commented):
This seems like a good idea.  Is it worth it to print a SHA to STDOUT?

danielpcox:
`git submodule update` prints the SHA when it changes the branch, which I think is enough.

PR #189, opened 2022-10-03
Upload binary upon release

dontlaugh:
Introduce an export_binary function and call it from build steps.
Embed the build number in the artifact's name.

The release step downloads the artifact and repackages it into
a tarball. Then the tarball is pushed to Nexus and Artifactory.


dontlaugh:
20221003_19h22m50s_grim

jack-bischoff (approved):

PR #190, opened 2022-10-11
gmcli unit tests

critterjohnson:
Tests the `gmcli` package of the operator using the mock client and randomly generated object data. The actual CLI is never called - this is tested elsewhere, and provided the CLI is adequately tested should not be required to verify the package works as intended.


danielpcox (approved):
Lookin' good, thanks Critter

PR #197, opened 2022-11-07
address vulnerabilities CVE-2022-27664 and CVE-2022-32149

moosecanswim:
[sc-30371]

Addresses CVEs by pinning `golang.org/x/net` and `golang.org/x/text` indirect dependencies to versions that have the fix.  While in here i also golang.org/x/crypto  version to `0.1.0` to address CVE-2022-30636.


go.mod
```diff
@@ -180,4 +180,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace github.com/go-git/go-git/v5 => github.com/QubitProducts/go-git/v5 v5.4.3-qubit
+replace (
```
joshuaswickirl, line 183: Is there a reason to have this second block as opposed to using the one on ln24?

go.mod
```diff
@@ -180,4 +180,8 @@ require (
...
+	github.com/go-git/go-git/v5 => github.com/QubitProducts/go-git/v5 v5.4.3-qubit
+	golang.org/x/net => golang.org/x/net v0.1.0
+	golang.org/x/text => golang.org/x/text v0.3.8
```
joshuaswickirl, line 186: x/text v0.4.0 is a dep of x/net. We don't need to declare text here again.

joshuaswickirl (changes requested):

clbx (approved):
lgtm

PR #200, opened 2022-11-29
Add namespace prefixing to sidecar injection

jack-bischoff:
Kubernetes only enforces resource name uniquness within a namespace.
The operator used strictly the resource name when injecting the
sidecar. This could cause two sidecars to announce themselves
as the same service when in fact the are in different namespaces
likely performing different missions. As a result, the control plane
would pass the same configuration to both.

The solution here is to prefix the resource name with the namespace
the resource is deployed in. For example with ns foo and resource
name bar, the resulting proxy name (XDS_CLUSTER) would be foo-bar.
This value is unique within the cluster.

This PR ensures the `XDS_CLUSTER` name and the pod label values include the `namespace-` prefix during sidecar injection and during optional configuring, ensures the resulting `Proxy.Name` field contains the prefixed name. Additionally it fixes the tests to accommodate that change and the other changes to greymatter-core.


alecholmez:
Thanks for spending the time to fix the tests and update the submodule. That was much needed

alecholmez:
I've tested this and its good to go.

alecholmez (approved):

alecholmez:
For some reason github thinks this is wip. Merging with admin privs.

PR #206, opened 2022-12-21
update the fips pipeline

moosecanswim:
[sc-32661]

This pr: 
- continues the compartmentalization of the pipeline
- builds *non prod* images in addition to generic artifats
- pushes branch and commit artifacts to dev-generic and dev-oci
- limits the integration test so they do not run when on a branch without a pr
- when a pr is opened the build will be run again and the pipeline will trigger integration tests
- on merge to main `latest` tagged items are pushed
- on tag of a commit the artifacts will labeled and pushed

- branch builds are about 2 min
- pr builds are 7+ min (integration tests)
- tagging and builds are seconds
- tagging build has a block for user acknowledgment 

> Will need to come back and add UBI images to build (currently ubuntu)


moosecanswim:
@danielpcox @dontlaugh  notifying yall since you did a bunch of work on the integration tests and i changed them up a bit.

dontlaugh:
As you probably know, the primary reason we went with generated pipelines here was because of the fact that - on startup - the transient k3s VM accepts its targeting rules from AWS ec2 tags. Traditionally we've tagged the EC2 with a build number and a pipeline slug (short name), ensuring that the agent that starts will match the specific targeting rules of the generated VM.

Launching pre-baked VMs is fast, but I'd be open to other approaches that accomplish roughly the same thing. Requirements are, roughly

1. We need an api to hit to launch stuff, webhook-style, likely from the public internet (relay.sh was what we started on but we need to move off that)
2. We need to be able to prevent other builds from running on the K3s VM.
3. We need VMs to get cleaned up, and a way to prevent that from happening if/when we want to ssh in and inspect a failed build

The so-called "launch" prototype is a daemon that I wrote that lives on an EC2 and - if asked - can launch and tear down VMs like this. But it's just a prototype on a single VM.

In the context of our migration to Azure, we'll want to figure out how Packer VM images work with that API, but the whole orchestration piece around this (launching on demand, teardown after ~3 hours).

It's totally possible to use something like buildkite schedules to drive this, if you can make the `az` command line queries work, but there are also benefits to writing a stateful service that can just centralize all the logic, auth, and logging. Or we buy Ansible Tower, etc.

joshuaswickirl (approved):

joshuaswickirl:
That's a clever solution. I think we should just stub/mock dependent services and use contract testing. The test harness would be reusable. We could then push most of the tests into unit tests.

We could do this with the acceptance tests too. Or use the same concept but at the container level.

joshuaswickirl:
These two categories for testing are all we need. If we wanted one or two canned smoke tests we should build that into the product and expose it to a curl /status or something.

PR #212, opened 2023-01-17
Support configurable CA bundle and no-verify option

danielpcox:
[sc-34205]

Associated docs PR: https://github.com/greymatter-io/gm-docs/pull/178


main.go
```diff
@@ -120,8 +120,12 @@ func run() error {
...
 	syncOpts = append(syncOpts, gitops.WithSSHInfo(syncSSHKeyPath, syncSSHKeyPassword))
 	syncOpts = append(syncOpts, gitops.WithRepoInfo(syncRepo, syncBranch, syncTag))
-	syncOpts = append(syncOpts, gitops.WithBasicAuthCredentials(
-		"GITOPS_BASICAUTH_USERNAME", "GITOPS_BASICAUTH_PASSWORD"))
+	// env var names chosen to match CLI/sync's
+	syncOpts = append(syncOpts, gitops.WithHTTPSGitOpsTargetOptions(
+		"GREYMATTER_GIT_USER",
+		"GREYMATTER_GIT_PASSWORD",
+		"GREYMATTER_GIT_REMOTE_CA",
+		"GREYMATTER_GIT_TLS_SKIP_VERIFY"))
```
dontlaugh, line 128: How does this work? Are these strings inputs to environment lookups? Or just placeholders that need to be replaced?

danielpcox: They are the environment variable names for those values. I opted to inject them rather than hide them inside the function that does the lookup, so they can be in main.go.

PR #216, opened 2023-01-23
integration tests are done on PR and on main builds

moosecanswim:
On friday with RC8 we were talking with daniel and realized that we should be including the integration tests on main branch builds.


joshuaswickirl (approved):

danielpcox:
That Daniel. Always stirring up trouble.

dontlaugh:
Any blocker here?

PR #219, opened 2023-01-24
Fix live upgrades to e.g. spire and edge tls

danielpcox:
[sc-34456]


pkg/gitops/state_backup.go
```diff
@@ -164,7 +164,7 @@ func NewSyncState(ctx context.Context, defaults cuemodule.Defaults) *SyncState {
 			DB:         defaults.RedisDB,
 			Username:   defaults.RedisUsername,
 			Password:   defaults.RedisPassword,
-			MaxRetries: -1,
```
alecholmez, line 167: Why this change?

pkg/mesh_install/install.go
```diff
@@ -1,10 +1,11 @@
 package mesh_install
 
 import (
+	"github.com/greymatter-io/operator/pkg/gmapi"
```
alecholmez, line 4: One helpful tip:

If you configure your editor to use `goimports` for formatting it'll organize your import list for you in a special order.

alecholmez (commented):
Just a question and a tip

pkg/mesh_install/install.go
dontlaugh: In Goland: Code > Optimize Imports

Hopefully the same as `goimports`;

pkg/gitops/state_backup.go
danielpcox: Just a robustness improvement. -1 is "no retry on connection failure", and after I had a look at the default backoff policy I decided this is good.

pkg/mesh_install/install.go
danielpcox: Optimize Imports seems to only remove unused imports, which already happens on save. I hope this one isn't a blocker, 'cause I don't care what order the imports are in?

danielpcox: Actually, I think I just figured it out. I can change the autoformatter to goimports instead of gofmt

danielpcox: And now my code is _so much better, wow_. JK.

danielpcox: Sorry, this would sound funny in person, instead of like I'm being a complete jerk.

pkg/gitops/state_backup.go
alecholmez: Gotcha. Was just curious where the arbitrary 10 retires came from.

alecholmez (approved):
I was able to test this and verify that it works.

alecholmez (changes requested):
Can we add a changelog update?

danielpcox:
Done

alecholmez (approved):

alecholmez (dismissed):

alecholmez (dismissed):

PR #222, opened 2023-02-07
Sleep between greymatter and k8s updates

danielpcox:
[sc-35361]

This PR just introduces a 5-second hard-coded sleep between the greymatter config update and the k8s config updates, so that if the k8s update bounces Control API and/or Redis, Control API won't lose parts of the config. There are other ways to do this, but they involve high-effort work in Control API that would be obviated by 2.0 (when we'll store this state in etcd).

I propose we simply sleep for >2 x the Control API backup ticker time (1 second). In practice, the CUE load still dominates the GitOps update time, so 5 seconds is hardly noticeable.

I have tested this with and without the sleep, and 5 seconds is sufficient and then some to solve the problem reliably.


alecholmez (approved):
I think this is fine. We'll want to note what we can actually do in control for 2.x to fix this problem. Making things reactive will help improve system speed across the board

PR #230, opened 2023-02-10
refactor v1alpha1 package name to meshv1

tilleryd:
There was some package naming cruft in operator that we didn't want to touch a few months ago. This cleans that up by renaming `v1alpha1` to `meshv1`.

See corresponding greymatter-core PR for more details https://github.com/greymatter-io/greymatter-core/pull/182.


PROJECT
```diff
@@ -12,8 +12,8 @@ resources:
     namespaced: true
   domain: greymatter.io
   kind: Mesh
-  path: github.com/greymatter-io/operator/api/v1alpha1
-  version: v1alpha1
+  path: github.com/greymatter-io/operator/api/meshv1
+  version: meshv1
```
dontlaugh, line 16: Is this PROJECT file still in use?

api/meshv1/groupversion_info.go
```diff
@@ -26,7 +26,7 @@ import (
 
 var (
 	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{Group: "greymatter.io", Version: "v1alpha1"}
+	GroupVersion = schema.GroupVersion{Group: "greymatter.io", Version: "meshv1"}
```
dontlaugh, line 29: I dont think this is valid according to Kubernetes' API versioning scheme.

https://book.kubebuilder.io/cronjob-tutorial/gvks.html#err-but-whats-that-scheme-thing

I need to find a better doc.

None of this matters, really, because as far as I know we aren't using CRDs anymore, which means we aren't actually creating Kubernetes API extensions. This never got cleaned up.

dontlaugh: In practical terms, Version here (I believe) should be one of the common generic strings you see in K8s yaml

* v1
* v1beta1
* v2
* etc...

dontlaugh: looks like you can _sort of_ use whatever you like. The parser tries to be smart. Note that a plain `v1` will take priority over something labeled `mesh1`, though.

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority

Of course, again, this might be all moot, since we aren't using CRDs?

danielpcox: Yeah, it's moot because it's no longer a CRD.

PROJECT
danielpcox: Not at all. We could delete this.

danielpcox (dismissed):
Good to go, though if you want to delete the PROJECT file, I'll re-approve.

PROJECT
tilleryd: Done

danielpcox (approved):
Reapproving

PR #231, opened 2023-02-10
On tag push to ci-staging-* in addition to dev-oci

moosecanswim:
[sc-35803]


.buildkite/tag.yaml
```diff
@@ -9,8 +9,11 @@
     env:
       JF_SRC_REGISTRY: dev-generic
       JF_DEST_REGISTRY: dev-generic
```
joshuaswickirl, line 11: Do we still need to push the tags to dev too?

joshuaswickirl (approved):

joshuaswickirl (approved):

PR #234, opened 2023-02-22
GMP-315 No reconciliation on replicas change

danielpcox:
GMP-315

Jon and the TIO team discovered that scaling sidecar-injected services up and down caused _all_ associated pods to restart. This is because `replicas` is under `spec`, and `spec` changes cause the generation of a K8s resource to advance, which is the condition for Operator reconciliation to notice and enforce its own changes.  This is undesirable, so this PR (tested in a kOps cluster) checks if the change affects replicas before deciding whether to enforce.


jimgaspari (approved):
Changes worked in testing

PR #238, opened 2023-03-01
Add robust update script based on gm-tools/scripts/update

danielpcox:
## How to test

Run `scripts/update` from anywhere, and it should do all of the steps necessary to update the `main` branch of the operator, and then return you to your original state, whether or not files were changed, and whether or not you were on `main` to begin with.


dontlaugh:
I tagged prodops. If we develop more sophisticated tools for dependency management, they should be shared and collectively maintained.

dontlaugh (commented):
Can you try another rebuild on this?

joshuaswickirl:
Do you want to merge this? We're going to be working in these files in https://github.com/greymatter-io/operator/pull/244 and don't want to hand you conflicts.

joshuaswickirl (approved):
Cool. Why is the temp branch needed?

danielpcox:
I'm rebuilding, but I guess I'll merge even if the build doesn't succeed 'cause the PR isn't the reason the build is failing.

The temp branch is necessary because you can't do `git fetch origin main:main` if you're already on main.

danielpcox:
Build failed again. Merging anyway.

PR #239, opened 2023-03-01
use ubi9

moosecanswim:
Fixes the build issue with ubi 9 minimal.  Wondering if microdnf installed tar with an issue since nothing changed on the image (last updated 2/7/23).  Nothing in the dockerfile or scripts should have impacted this.

This brings up the question.  Should we host builder images with this stuff baked in?  We should investigate if that the issue.  If it is then maybe thats something we should think about.


moosecanswim:
From breaking pipeline build:
```
tar-2:1.34-6.el9_1.x86_64                        ubi-9-baseos-rpms    910.7\xc2\xa0kB
```

From old working build:
```
tar-2:1.34-5.el9.x86_64                        ubi-9-baseos-rpms    911.1\xc2\xa0kB
```

moosecanswim:
@joshuaswickirl when you get back on if you want to try minimal with a specified version that might make this a smaller change.

joshuaswickirl (approved):

PR #244, opened 2023-03-06
OPS-18 Replace relay

joshuaswickirl:
OPS-18

This PR does a number of things:
- replaces relay.sh integration tests w/ aks cluster
    - new pipeline components utilize the new kubernetes based build system
    - utilizes external secrets from vault instead of baked in value
    - defines build dependencies that the pipeline builder uses
    - cleans up all resources created
- rewrite of the tests to accommodate a real kubernetes cluster
- removal of unused scripts in `scripts/cibuild`


joshuaswickirl:
This is looking great so far. It's a shame we have to change the tests so much.

moosecanswim:
Part of what makes this take a long time is that we're waiting for the operator to be ready before running any tests.  we may want to consider lowering this since the greymatter mesh is ready even before the operator is "ready"

```
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /healthz
            port: 8081
            scheme: HTTP
          initialDelaySeconds: 120
          periodSeconds: 20
          successThreshold: 1
          timeoutSeconds: 1
```

.buildkite/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+docker build --platform linux/amd64 -t operator-custom-build-image:latest -f .buildkite/dockerfiles/integration-test-image .
+``` 
+
+> Before tagging make sure your image follows the naming convention outlined in <TODO:spec this out>
+
+To re-tag your local image:
+
+```console
+docker tag operator-custom-build-image:latest greymatter.jfrog.io/internal-oci/operator-build-image:v0.10
```
clbx, line 21: ``internal-oci`` or ``pipeline-oci``?

scripts/cibuild
```diff
@@ -13,244 +13,50 @@ cmd_bootstrap () {
...
-    _bundle_and_push "amd64" "linux" ${BUILDKITE_TAG:1}
-  else
-    echo "no BUILDKITE_TAG present"
-  fi
+  az aks create --resource-group ${az_group_name} \
+    --name ${az_cluster_name} \
+    --enable-managed-identity \
+    --node-count 2 \
+    --node-vm-size Standard_DS4_V2 \
+    --enable-cluster-autoscaler \
```
clbx, line 32: What's the benefit of using autoscaling if the cluster is very short lived and only tests the operator?

moosecanswim: Ah nice catch.  I was using that when i was trying to figure out the size of cluster i needed.  Will remove!

.buildkite/README.md
moosecanswim: I like `pipeline-oci`

moosecanswim: also gonna bump to 1.0

clbx (dismissed):

moosecanswim:
Spun off this PR in greymatter core https://github.com/greymatter-io/greymatter-core/pull/205

dontlaugh:
> i would like to come back to how we are using patches

I think the fact that we felt forced to use these in the first place signals that we need to do work on operator's interaction with greymatter-core. We might be able to provide operator with a little bit more direct config that picks/chooses/instantiates more pieces of the CUE tree.

I am looking at this now, albeit for a different reason, in the context of https://greymatter.atlassian.net/browse/GMP-431

dontlaugh (commented):
No blockers from me

moosecanswim:
@danielpcox was gonna respond in slack but then it became realy long

Yeah, Its a pretty big change.

When you boil it all down its the same code being sent to the the helper shell function (real k8s instead of k3s).  However to get to those theres a declarative MeshTest class.  This class provides the basics that should always be run when a test is run.  The tests can be expanded however they will always include the same base functionality.

The tests have not been running since relay.sh was shut down.  So that someone could look at test
```
    defaultTest=mt.MeshTest(
        name="default-test", 
        operator_namespace="gm-operator", 
        mesh_namespace="greymatter",
        elasticsearch_namespace=elasticsearch_dependency.namespace,
        init_patches=[
            "scripts/fixtures/gitops_setup.patch"
        ],
        buildkite_config=buildkite_config
    )
    tests.update({defaultTest.name: defaultTest})
```
and easily see “for default test we’re going to be deploying the operator into the gm-operator namespace, greymatter into the greymatter namespace, we’re pointing to an elasticsearch namespace and we’re implementing the gitops_setup patch”.  Similar for the spire test.  Different namespaces and more patches.  

Additionally the commands arguments per the test in script:
```
./scripts/test k8s_test_runner apple
Running from /Users/kylegallagher/Documents/work/decipher/operator/operator
Do k8s Test [apple]
Invalid K8s Test to run
Test Options:
    default-test
    spire-second-mesh-test
```

if i were to add a test:
```
    secondDefaultTest=mt.MeshTest(
        name="secondDefaultTest", 
        operator_namespace="gm-operator2", 
        mesh_namespace="greymatter2",
        elasticsearch_namespace=elasticsearch_dependency.namespace,
        init_patches=[
            "scripts/fixtures/gitops_setup.patch",
            "scripts/fixtures/second_operator.patch"
        ],
        buildkite_config=buildkite_config
    )
    tests.update({secondDefaultTest.name: secondDefaultTest}) 
```

then it becomes available for use
```
❯ ./scripts/test k8s_test_runner apple
Running from /Users/kylegallagher/Documents/work/decipher/operator/operator
Do k8s Test [apple]
Invalid K8s Test to run
Test Options:
    default-test
    secondDefaultTest
    spire-second-mesh-test
```

Theres still work to be done.  Id like to tie in dependency status and tidy up init work however i think this will make the test structure a lot more approachable for those with limited kubernetes experience.

joshuaswickirl (commented):
Seems that the API is creating the mesh tests but right now that's nested away in MeshTest.py. It'd be nice if these definitions were easier to find from the entry point. Also, not a huge fan of inheritance here. There isn't a real reason to couple the different tests.

Beyond that there seems to be a lack of cohesion, I found myself hopping around the 4 files quite a bit to understand what was going on. Some of this comes from the lack of meaning in some of the function names forcing me to read the implementation.

Overall, it gets the job done, so I'm good with it. We can iterate from here.

... but I'm the pr author so all you get is comments from me.

joshuaswickirl:
Seems like the scope got blown up a little here too. This PR is a bit big and the changes to the python overshadow the necessary changes for replacing relay. https://greymatter.atlassian.net/browse/OPS-18

danielpcox (approved):
Let's merge this and then evolve the system from there. IMO it is much more important to have the integration tests running at all than it is to have them running _in a particular way_. I'm really grateful that you guys got them working again.

That said, my major complaint is that the content of the integration tests themselves didn't have to change at all to replace Relay.sh, only the harness around them, and the changes in this PR reverse the philosophy. Originally the idea was that since lots of people already know how to _manually_ test operator PRs, let's make a very imperative, sequential set of integration tests that do things exactly the way people manually test, so that it's trivial to translate a manual test into an automated test - just run the shell scripts in `shell`. A few other helpers were provided to do common things like waiting for pods, and patch files are an annoying compromise to simulate editing files in the middle of a test, but crucially I didn't want to build a DSL that people would have to learn before they could write an integration test. I didn't want them to have to read the code in detail before they could do that. I think whatever brilliant declarative scheme I could come up with would prevent people from writing tests, so I eschewed obfuscation and wrote something unbeautiful but straightforward.

Anyway, all of _that_ said, I'd also rather encourage you to think for yourself than to try to guess what was in Daniel's head. Good work, and let's build on this. Maybe we can achieve both sets of goals.

scripts/lib/MeshTest.py
```diff
@@ -0,0 +1,158 @@
...
+        edge_host = h.get_svc_external_ip(service='edge', namespace=self.mesh_namespace)
+        edge_port = h.confirm_mesh_reachable(edge_host, port=10808)
+        return edge_host, edge_port
+    
+    def mesh_tests(self, edge_host, edge_port):
+        # Check pre deploy service has connection to it
+        h.wait_for_pod('greymatter.io/cluster=default-simple-server', 'default')  # confirm annotated Deployment runs
+        h.confirm_gitops_update(edge_host, edge_port)
+        # Check simple server has been registered with the mesh
+        h.curl(f'http://{edge_host}:{edge_port}/services/simple-server')  # confirm sidecar was injected and configured
```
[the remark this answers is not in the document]
danielpcox: Yes, `curl` retries a bunch of times by default and eventually raises. https://github.com/greymatter-io/operator/blob/replace-relay/scripts/lib/helpers.py#L99

PR #245, opened 2023-03-07
Remove priority class

joshuaswickirl:
See: OPS-26 https://greymatter.atlassian.net/browse/OPS-26


alecholmez:
This will need a core patch as well: https://github.com/greymatter-io/greymatter-core/blob/main/k8s/outputs/controlensemble.cue#L42

Can we get that in first so the commits line up for all this stuff?

joshuaswickirl:
Yeah that's here. https://github.com/greymatter-io/greymatter-core/pull/197

joshuaswickirl:
> This will need a core patch as well: https://github.com/greymatter-io/greymatter-core/blob/main/k8s/outputs/controlensemble.cue#L42
> 
> Can we get that in first so the commits line up for all this stuff?

Yeah, here's the pr: https://github.com/greymatter-io/greymatter-core/pull/197

moosecanswim (approved):
LGTM.  Tho we dont really need to remove it from the operator since the operator will only create the resource if greymatter core specifies it.

alecholmez (approved):

alecholmez:
I'd rather remove for now and then add back if a customer really needs this. Dead code = technical debt

PR #243, opened 2023-03-03
Refactor program startup

dontlaugh:
- Refactor program startup
- Centralize validation in Sync constructor

PR #246, opened 2023-03-07
Bugfix: Overwrite cueRoot if syncRepo is set

dontlaugh:
Without this fix, the operator will clone your custom branch successfully, but not actually use it to deploy.


dontlaugh:
This was introduced by #243

dontlaugh:
I tested this on Nixon; https://github.com/greymatter-io/nixon/blob/main/inputs.cue#L44-L45

It updated the catalog container via gitops.

PR #248, opened 2023-03-08
Fixing finalization

danielpcox:
Fixes GMP-364 by
a) always being careful to do reconciliation on deletion
b) comparing replicas counts after a pointer dereference

- Also bump the submodule to resynchronize the operator with greymatter-core (PriorityClasses were removed)
- Also updating one of the integration test patches so we can remove resource limits at will.

## How to test
- There is a test build of this at [greymatter.jfrog.io/internal-oci/greymatter-operator:finalizer-fix](http://greymatter.jfrog.io/internal-oci/greymatter-operator:finalizer-fix)
- To test, deploy a cluster using that version of the operator, deploy a Deployment in a watched namespace with `greymatter.io/inject-sidecar-to: "9090"` like so:
```
  # workload.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: fake-service
  annotations:
    greymatter.io/inject-sidecar-to: "9090"
spec:
  selector:
    matchLabels:
      app: fake-service
  template:
    metadata:
      labels:
        app: fake-service
    spec:
      containers:
        - name: fake-service
          image: greymatter.jfrog.io/internal-oci/fake-service:v0.24.2
```
- Wait for it to be deployed
- Then delete it. Previously, the delete would hang forever because reconciliation wasn't happening on deletes.

Testing that the ignore-replicas-changes functionality still works
- Redeploy the Deployment above
- Wait for it to come up with a sidecar
- Scale it up to 5 nodes and wait for it to happen
- Scale it down to 2, and notice that the first two do not restart.
- Scale down to 1 and notice that the oldest one does not restart.
- Do this a few more times to be sure that it doesn't restart all replicas whenever _any_ replicas are added or removed.


joel-epstein:
I am testing this now.

joel-epstein (approved):
Deployed greymatter-core from `main` and tested the finalization fix with a simple GSL project/service.  Permuted through scaling up/down and verified configs were still applied.  Checked Dashboard as well.

PR #258, opened 2023-03-28
Ordered k8s apply

moosecanswim:
OPS-58


pkg/mesh_install/install.go
```diff
@@ -155,3 +157,80 @@ func (i *Installer) triggerInjectedSidecarReconciliation() {
 		}
 	}
 }
+
+type KindSortOrder []string
+
+var InstallOrder KindSortOrder = []string{
```
alecholmez, line 163: You've created a type mask so you don't need to redeclare in the literal syntax
```suggestion
var InstallOrder = KindSortOrder{
```

pkg/mesh_install/install.go
```diff
@@ -155,3 +157,80 @@ func (i *Installer) triggerInjectedSidecarReconciliation() {
 		}
 	}
 }
+
+type KindSortOrder []string
```
alecholmez, line 161: Doc string?

pkg/mesh_install/install.go
alecholmez: Also Doc string since it's exported

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## vNEXT
 
+### Changed
```
alecholmez, line 5: Should this be an add? The comment notes an `Added` feature. I'd reword to match the semantics of the changelog header

pkg/mesh_install/install.go
```diff
@@ -155,3 +157,80 @@ func (i *Installer) triggerInjectedSidecarReconciliation() {
...
+	"Job",
+	"CronJob",
+	"IngressClass",
+	"Ingress",
+	"APIService",
+}
+
+func sortManifestsByKind(manifests []client.Object, ordering KindSortOrder) []client.Object {
+	sort.SliceStable(manifests, func(i, j int) bool {
+		return lessByKind(manifests[i], manifests[j], manifests[i].GetObjectKind().GroupVersionKind().Kind, manifests[j].GetObjectKind().GroupVersionKind().Kind, ordering)
```
alecholmez, line 205: This line is super long, we can break it a tiny bit?

danielpcox (approved):
Heck yes

pkg/mesh_install/install.go
```diff
@@ -155,3 +157,80 @@ func (i *Installer) triggerInjectedSidecarReconciliation() {
...
+	})
+
+	return manifests
+}
+
+func lessByKind(a interface{}, b interface{}, kindA string, kindB string, o KindSortOrder) bool {
```
alecholmez, line 211: Top level scoped variables and parameters should generally be the most descriptive in Go. Short variable names are intended for tightly scoped blocks, like `err` or `i` in a tiny for loop.
```suggestion
func lessByKind(a interface{}, b interface{}, kindA string, kindB string, order KindSortOrder) bool {
```

alecholmez: Also do `a` and `b` need to be `interface{}`? Can they be `client.Object` since that's what gets passed in from the wrapper sort function?

alecholmez (changes requested):

CHANGELOG.md
moosecanswim: Not really a feature for users.  It modifies how k8s stuff are applied.

alecholmez: I'm looking for something like:
```
- Changed the k8s manifest installation process to specify an order of priority given the resource kind
```

alecholmez:
Also FWIW I think we need to drop the helm license in this source code too. This is their implementation so we should cover our bases legally. Maybe it's worth breaking this out into a separate file so we can drop the helm license at the top of the file? It's apache 2 which is good just need to make sure we give credit.

CHANGELOG.md
moosecanswim: Updated comment and moved into fixed section.

pkg/mesh_install/install.go
moosecanswim: broke it up into more lines

pkg/mesh_install/install.go
moosecanswim: didnte even need that.  just used the string kind cause thats all we were using

moosecanswim:
Looking at this it looks like we just need a notice file that sites helm and states that we derived code from what they have.  https://www.apache.org/licenses/LICENSE-2.0.html#redistribution

alecholmez:
> Looking at this it looks like we just need a notice file that sites helm and states that we derived code from what they have. https://www.apache.org/licenses/LICENSE-2.0.html#redistribution

Beautiful, that works for me. As long as we give credit here I'm good with that

PR #260, opened 2023-03-29
Rearange the k8s objects before running the apply

moosecanswim:
OPS-58

The way that the operator evaluates cue does not enforce an order.  In previous logs we’ve seen that service accounts are not created before pods that are expecting the service accounts.  To account for this we implemented a system similar to helm that sorts the manifests before applying them.


pkg/mesh_install/install.go
```diff
@@ -155,3 +157,85 @@ func (i *Installer) triggerInjectedSidecarReconciliation() {
 		}
 	}
 }
+
+type kindSortOrder []string
+
+var installOrder kindSortOrder = []string{
```
alecholmez, line 163:
```suggestion
var installOrder = kindSortOrder{
```

alecholmez (changes requested):
One last tiny thing then this is good to merge.

alecholmez:
We you also update the PR description with WHY this change is necessary? The ticket outlines why this is good but the git history should also denote why the fix is necessary.

alecholmez (approved):
Approved thank you!

